PR #1, opened 2019-03-15
Prepare initial third party services

Fryguy:
Prepare initial third party services


.travis.yml
```diff
@@ -1,20 +1,19 @@
...
-  global:
-  - DATABASE_URL="postgresql://postgres:@localhost:5432/sources_test?encoding=utf8&pool=5&wait_timeout=5"
+  postgresql: '10'
 before_install:
-- gem update --system
+- 'echo ''gem: --no-ri --no-rdoc --no-document'' > ~/.gemrc'
 - gem install bundler
+- DATABASE_URL="postgresql://postgres:@localhost:5432/sources_api_test?encoding=utf8&pool=5&wait_timeout=5"
 before_script:
-- cp ./v2_key.dev ./v2_key
```
agrare: We still need this for specs to pass

.travis.yml
```diff
@@ -1,20 +1,19 @@
...
-  postgresql: '9.5'
-env:
-  global:
-  - DATABASE_URL="postgresql://postgres:@localhost:5432/sources_test?encoding=utf8&pool=5&wait_timeout=5"
+  postgresql: '10'
 before_install:
-- gem update --system
+- 'echo ''gem: --no-ri --no-rdoc --no-document'' > ~/.gemrc'
 - gem install bundler
+- DATABASE_URL="postgresql://postgres:@localhost:5432/sources_api_test?encoding=utf8&pool=5&wait_timeout=5"
```
agrare: Should this be in `env.global`?

Fryguy: no?  I don't know what you mean

agrare: The tests are failing when looking for a config/database.yml, I assume because there is no DATABASE_URL, possibly because this should be in the env global section rather than in before_install.

Fryguy: Oh you mean in the section above?  No, it's not necessary...you only really need the env section if you plan on testing a matrix where the env changes actually emit multiple jobs.  See https://docs.travis-ci.com/user/build-matrix/

Fryguy: This is how I set it up in all repos.  Not sure why this is different.  Let's discuss

Fryguy: ok fixed

PR #14, opened 2019-04-17
Move API from v0.1 to v1.0

agrare:
Since no one is officially using the v0.1 API lets move over to v1.0

https://github.com/ManageIQ/sources-api-client-ruby/pull/1


agrare:
cc @abellotti @gtanzillo

PR #18, opened 2019-04-25
Multilanguage client generator

slemrmartin:
rake task for generating client in more languages, defaulting ruby

Same PR as https://github.com/ManageIQ/topological_inventory-api/pull/187


slemrmartin:
cc @agrare, @abellotti

PR #19, opened 2019-04-26
Update source product_name according to UX.

martinpovolny:
according to UX: https://docs.google.com/document/d/1cE_sG6IdBW-cHCODjplbWzDldHDhsbW6Xyn--XQqNxQ/edit?ts=5cc0b425#heading=h.wqa0bh1el605


carbonin:
This will need a migration, right?

agrare:
@carbonin we re-seed on deployment so this should be sufficient

carbonin:
Ah, okay. Great!

PR #20, opened 2019-05-03
Entitlement enforcement

gtanzillo:
This change will enforce entitlements if the "entitlements" key is present in the x-rh-identity header.

If the "entitlements" key is present, the user must be entitled for either "insights" or "hybrid_cloud". I neither is entitled a 403 response is returned

Copied from https://github.com/ManageIQ/topological_inventory-api/pull/189

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -45,14 +45,17 @@ class ApplicationController < ActionController::API
     ManageIQ::API::Common::Request.with_request(request) do |current|
       begin
         if Tenant.tenancy_enabled? && current.required_auth?
-          tenant = Tenant.find_or_create_by(:external_tenant => current.user.tenant)
+          raise ManageIQ::API::Common::EntitlementError unless request_is_entitled?(current.entitlement)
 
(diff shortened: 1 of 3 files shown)


bdunne:
@gtanzillo Is this ready to go?  Can you address the rubocop comments?

app/controllers/application_controller.rb
```diff
@@ -45,14 +45,17 @@ def with_current_request
...
 
+          tenant = Tenant.find_or_create_by(:external_tenant => current.user.tenant)
           ActsAsTenant.with_tenant(tenant) { yield }
         else
           ActsAsTenant.without_tenant { yield }
         end
       rescue KeyError, ManageIQ::API::Common::IdentityError
         render :json => { :message => 'Unauthorized' }, :status => :unauthorized
+      rescue ManageIQ::API::Common::EntitlementError
+        render :json => { :message => 'Forbidden' }, :status => :forbidden
```
bdunne: Sorry for just noticing this, but shouldn't this method be rendering error documents instead?

PR #21, opened 2019-05-08
Add support to generically handle ActiveRecord::NotNullViolation exception

abellotti:
Fixes https://projects.engineering.redhat.com/browse/TPINVTRY-222

Add support to generically handle ActiveRecord::NotNullViolation exceptions and return a proper error message instead of the 500 internal error.

Massage the exception message to get the following example:

{
  "errors": [
    {
      "status": 400,
      "detail": "Missing parameter - null value in column \"name\" violates not-null constraint"
    }
  ]
}

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,12 @@ class ApplicationController < ActionController::API
     render :json => exception.error_document.to_h, :status => exception.error_document.status
   end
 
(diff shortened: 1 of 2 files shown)


abellotti:
/cc @agrare or @gtanzillo please take a 👀 

tried to handle this generically, we could have just raised an error in the sources controller for missing name, though this approach handles any such exception.  Let me know if this is ok with you.  Thanks.

PR #22, opened 2019-05-08
Assures that name specified is specified and not blank

abellotti:
This fixes: https://projects.engineering.redhat.com/browse/TPINVTRY-223

Added model validator to assure that name is specified and is non-blank,
added rescue for ActiveRecord::RecordInvalid and appropriate spec for the empty name scenario.


abellotti:
ping @agrare please take a look when you get a chance, I just want to make sure JIRA is correct (name specified cannot be blank), and fix ok.  Let me know if anything needs to be updated, Thanks.

PR #24, opened 2019-05-13
Fix for POST create openapi JSON signatures

abellotti:
Response signatures like:

             "content": {
               "application/json": {
                 "schema": {
                   "type": "object",
                   "items": {
                     "$ref": "#/components/schemas/Source"
                   }
                 }
               }
             }

  Should just be for simple json objects and not arrays:

             "content": {
               "application/json": {
                 "schema": {
                   "$ref": "#/components/schemas/Source"
                 }
               }
             }

### DIFF
--- a/lib/tasks/openapi_generate.rake
+++ b/lib/tasks/openapi_generate.rake
@@ -273,10 +273,7 @@ class OpenapiGenerator
           "description" => "#{klass_name} creation successful",
           "content"     => {
             "application/json" => {
-              "schema" => {
-                "type"  => "object",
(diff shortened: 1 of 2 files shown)


agrare:
:+1: thanks @abellotti 
Was this WIP just because of the spec failure or do you want to do more testing?

abellotti:
Just waiting to hear from Shalom from his tire kicking.  I'll then unwip here and create the same for the topo api.

abellotti:
Thanks @agrare for merging, equivalent (openapi:generate task only) PR in https://github.com/ManageIQ/topological_inventory-api/pull/193 for topo.

PR #27, opened 2019-05-21
Set the auth_type for authentications

agrare:
Set the authentication.auth_type for newly created authentication
records.


agrare:
Hey @martinpovolny can you confirm if this will set the auth_type correctly in the UI?

PR #28, opened 2019-05-22
Fix Net undefined exception when generating client

agrare:
```
Sources API Version: 1.0
rake aborted!
NameError: uninitialized constant Net
/home/agrare/src/topological_inventory/sources-api/lib/tasks/client_generate.rake:30:in `generator_cli_jar'
/home/agrare/src/topological_inventory/sources-api/lib/tasks/client_generate.rake:57:in `generate_client'
/home/agrare/src/topological_inventory/sources-api/lib/tasks/client_generate.rake:74:in `block (2 levels) in <top (required)>'
/home/agrare/.gem/gems/rake-12.3.2/exe/rake:27:in `<top (required)>'
Tasks: TOP => client:generate
(See full trace by running task with --trace)
```

### DIFF
--- a/lib/tasks/client_generate.rake
+++ b/lib/tasks/client_generate.rake
@@ -26,6 +26,8 @@ class ClientGenerator
     @generator_cli_jar ||= begin
       jar_path = Pathname.new(Rails.root.join("public/doc/openapi-generator-cli-#{VERSION}.jar"))
       unless File.exist?(jar_path) && File.size(jar_path).positive?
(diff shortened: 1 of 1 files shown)


carbonin:
Did we write this or did this come from somewhere else? We'll probably need to make this change everywhere, right?

agrare:
@carbonin yeah I'm changing it in topological_inventory-api also, I think those are the only two but I'll check catalog as well

PR #29, opened 2019-05-22
Add rubocop/rubocop-performance to fix CodeClimate

agrare:


agrare:
@djberg96 is there anything else needed to get codeclimate to pass?

djberg96:
@agrare In theory that should do it. :)

PR #31, opened 2019-05-24
Add in the x-rh-identity headers to the event

agrare:
When we raise an even on sources crud operations add the
current_forwardable headers to the event headers so that dependent
services know e.g. the tenant/request id that led to the change.

Depends on: ~~https://github.com/ManageIQ/manageiq-messaging/pull/46~~

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 gem 'jbuilder',             '~> 2.0'
 gem 'json-schema',          '~> 2.8'
 gem 'manageiq-loggers',     '~> 0.1'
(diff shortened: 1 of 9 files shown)


infinitewarp:
Any progress on this? cloudigrade aka Cloud Meter will require this information, and we're blocked until it's available.

carbonin:
@agrare is on PTO for a while, but it looks like this is just missing a release of a dependent gem ... @gtanzillo can we get someone to release manageiq-messaging so that this can go green?

carbonin:
Pinged @bzwei about the gem release issue. Looks like the new version was tagged but not pushed to rubygems.org somehow ...

agrare:
@carbonin I just pushed the release to rubygems

PR #33, opened 2019-05-29
Create Source with given uid

slemrmartin:
For sync between Cloudforms and Sources I need create Source with first payload. 
It means I'll provide UID from CF: `ExtManagementSystem.guid`. 
So I need to create Source with given uid.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,6 +19,11 @@ class ApplicationController < ActionController::API
     render :json => error_document.to_h, :status => :bad_request
   end
 
+  rescue_from ActiveRecord::RecordNotUnique do |_exception|
+    error_document = ManageIQ::API::Common::ErrorDocument.new.add(400, "Record not unique")
+    render :json => error_document.to_h, :status => :bad_request
+  end
+
   rescue_from ManageIQ::API::Common::Filter::Error do |exception|
     render :json => exception.error_document.to_h, :status => exception.error_document.status
   end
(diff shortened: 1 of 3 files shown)


slemrmartin:
@abellotti I want something like find_or_create (I don't want to do two requests - 1. get, 2. create). 
Is the right way in handle HTTP 500 error, or is there a better way?

cc @agrare

slemrmartin:
@abellotti as I'm looking to specs, it should return status 400 with `ManageIQ::API::Common::ErrorDocument.new.add(400, "Invalid parameter - Validation failed: UID Not Unique")`, right?

agrare:
Looks good to me, @abellotti how does the error doc look to you?

slemrmartin:
I talked to UI guys and it seems I cannot avoid 2 requests...which means this is useless (but `rescue_from` could be handled in all cases)

app/controllers/application_controller.rb
```diff
@@ -19,6 +19,11 @@ class ApplicationController < ActionController::API
     render :json => error_document.to_h, :status => :bad_request
   end
 
+  rescue_from ActiveRecord::RecordNotUnique do |_exception|
+    error_document = ManageIQ::API::Common::ErrorDocument.new.add(400, "Record not unique")
```
abellotti: can we append a bit more info here as to what was the exception, i.e. to see which field was the issue.
Maybe do something similar to line 17 above and return something like "Record not unique - #{exception_msg}"
might be more useful for caller.

slemrmartin: We can, but there is no nice format (error doesn't contain which attribute is wrong, but what index is wrong), because it doesn't have active_record validation but database validation

agrare:
Okay going to merge this, if it ends up being possible to add more info to the error we can do that in a follow-up

PR #35, opened 2019-06-11
Updated the GraphQL specs to reflect the simplified query signature.

abellotti:
- Updated the GraphQL specs to reflect the simplified query signature.


abellotti:
WIP - depends on https://github.com/ManageIQ/manageiq-api-common/pull/78

PR #36, opened 2019-06-12
Changed openapi ID regexp pattern according to validator

slemrmartin:
openapi validator needs ID regexp pattern without "/" at the start and at the end.

* [x] **depends on**: https://github.com/ManageIQ/manageiq-api-common/pull/79

cc @abellotti

PR #37, opened 2019-06-13
POST/PATCH validations against openapi.json

slemrmartin:
Create/update requests are validated against openapi.json schema for parameter format. 
It checks only existing properties, unknown properties are checked yet.

---

* [ ] **depends on**: #36 
* [x] **depends on**: https://github.com/ManageIQ/manageiq-api-common/pull/80
* **Issue**: https://github.com/ManageIQ/manageiq-api-common/issues/67

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -64,14 +64,11 @@ class ApplicationController < ActionController::API
     return unless request.post? || request.patch?
 
     api_version = self.class.send(:api_version)[1..-1].sub(/x/, ".")
-    request_path = request.path.split(api_version)[1]
-    payload = body_params.as_json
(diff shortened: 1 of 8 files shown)


slemrmartin:
@gtanzillo @abellotti please review

slemrmartin:
@bdunne please review

app/controllers/application_controller.rb
```diff
@@ -57,12 +58,32 @@ def with_current_request
     end
   end
 
+  # Validates against openapi.json
+  # - only for HTTP POST/PATCH
```
Ladas: @slemrmartin the validation validates also parameters, etc. so it should be used for GET request too

PR #41, opened 2019-06-21
Update README stating dependency on ruby 2.4

agrare:
With less than ruby 2.4 you get the following error:

```
NoMethodError (undefined method `named_captures' for #<MatchData:0x00007ff9a3ad4928>):
app/controllers/application_controller.rb:205:in `request_path_parts'
app/controllers/application_controller.rb:209:in `subcollection?'
app/controllers/api/v1/mixins/index_mixin.rb:24:in `raise_unless_primary_instance_exists'
app/controllers/api/v1/mixins/index_mixin.rb:6:in `index'
app/controllers/application_controller.rb:51:in `block (2 levels) in with_current_request'
app/controllers/application_controller.rb:51:in `block in with_current_request'
app/controllers/application_controller.rb:46:in `with_current_request'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 This project exposes an API for accessing objects living in the Sources Service database
 
 ## Prerequisites
-You need to install ruby >= 2.2.2 and run:
+You need to install ruby >= 2.4 and run:
 
 ```
 bundle install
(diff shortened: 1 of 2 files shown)


agrare:
Thanks @carbonin

PR #42, opened 2019-06-21
Updated the openapi generator to use the new graphql description

abellotti:
- Moving the GraphQL Response into a separate schema definition as the client generator was getting confused with the inline response definition.


abellotti:
WIP - depends on https://github.com/ManageIQ/manageiq-api-common/pull/83

abellotti:
@miq-bot assign @bdunne

PR #44, opened 2019-06-24
Add handling of required properties

agrare:
The openapi generator wasn't handling properly if attributes were
required.  This will mark an attribute as required if it isn't in the
set of read-only classes/attributes and if it is not-nullable and
doesn't have a db level default.

### DIFF
--- a/lib/tasks/openapi_generate.rake
+++ b/lib/tasks/openapi_generate.rake
@@ -220,10 +220,20 @@ class OpenapiGenerator
       if GENERATOR_READ_ONLY_DEFINITIONS.include?(klass_name) || GENERATOR_READ_ONLY_ATTRIBUTES.include?(key.to_sym)
         # Everything under providers data is read only for now
         properties_value['readOnly'] = true
+      else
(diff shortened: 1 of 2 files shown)


agrare:
cc @abellotti I can't add you as a reviewer but can you take a look too?

agrare:
Still need to handle if there is a default on the model e.g. Sources.uid.  Best I could come up with was `model.new.columns_hash[key].present?` but I'm open to other thoughts

lib/tasks/openapi_generate.rake
```diff
@@ -220,12 +220,22 @@ class OpenapiGenerator
...
       properties_value.sort.to_h
     end
   end
 
+  def required?(_klass_name, model, key, value)
+    nullable = value.null && model.validators_on(key).none? { |v| v.kind == :presence }
+    has_default = value.default.present?
+
+    # If an attribute isn't nullable and has no default then it is required
+    !nullable && !has_default
```
abellotti: Nice !! 👍

PR #45, opened 2019-06-25
Required for schemas is a separate section

agrare:
For paths, required is just an attribute on the property but for schemas it is its own section at the same level as "properties".  This was causing invalid openapi schemas.


agrare:
Hm seems we can't add required fields and not break PATCH api calls which don't have to send all fields, going to close this and try to come up with a better solution.

PR #49, opened 2019-07-02
Undo required attributes for objects

agrare:
Objects don't have required attribute similar to how endpoints are.
This causes invalid swagger specs.  Introduced by https://github.com/ManageIQ/sources-api/pull/44


agrare:
Thanks @bdunne !

PR #50, opened 2019-07-02
GraphQL specs were not running with rspec

abellotti:
- GraphQL specs were not running because of the name of the spec file
- Moved the skip_before_action :validate_request in the graphql controller in the proper section.


abellotti:
ping @agrare

PR #52, opened 2019-07-15
Do not include `tenant` in any API responses

gtanzillo:
https://projects.engineering.redhat.com/browse/TPINVTRY-416

### DIFF
--- a/spec/requests/api/graphql_controller_spec.rb
+++ b/spec/requests/api/graphql_controller_spec.rb
@@ -18,7 +18,6 @@ RSpec.describe("v1.0 - GraphQL") do
       post("/api/v1.0/graphql", :headers => headers, :params => graphql_source_query)
 
       expect(response.status).to eq(200)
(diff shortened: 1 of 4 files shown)


spec/requests/api/graphql_controller_spec.rb
```diff
@@ -18,7 +18,7 @@ def result_source_tenant(response_body)
       post("/api/v1.0/graphql", :headers => headers, :params => graphql_source_query)
 
       expect(response.status).to eq(200)
-      expect(result_source_tenant(response.body)).to match_array([tenant.external_tenant])
+      # expect(result_source_tenant(response.body)).to match_array([tenant.external_tenant])
```
gtanzillo: Commented out for now. Need to decide if this test is still needed now that tenant is no longer returned

agrare: @gtanzillo you want to delete this line?

PR #56, opened 2019-07-19
Reinstating tenant in API responses

gtanzillo:
Tenant (in this case account number) is necessary in API responses because apps and services subscribing to source
activity messages on kafka need a way to associate them with the specific account making the change.

https://projects.engineering.redhat.com/browse/TPINVTRY-416

### DIFF
--- a/app/models/concerns/tenancy_concern.rb
+++ b/app/models/concerns/tenancy_concern.rb
@@ -5,4 +5,10 @@ module TenancyConcern
     belongs_to :tenant
     acts_as_tenant :tenant
   end
+
+  def as_json(*)
+    super.except("tenant_id").tap do |hash|
+      hash["tenant"] = tenant.external_tenant
+    end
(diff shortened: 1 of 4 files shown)


Fryguy:
@gtanzillo If the code is the same as before can you commit it as a `git revert` instead?

gtanzillo:
> @gtanzillo If the code is the same as before can you commit it as a `git revert` instead?

@Fryguy, that's what I did but then amended the revert commit with an explanation

Fryguy:
Ah ok...usually a revert commit references the original commit...did you remove that when you ameded?

gtanzillo:
> Ah ok...usually a revert commit references the original commit...did you remove that when you ameded?

Yes, I did. I can add back

agrare:
@gtanzillo can you cherry-pick this to the stable branch?

PR #57, opened 2019-07-23
Normalize user name fields in json schema

gtanzillo:
Make user name fields in the json schema, used by the Sources UI to be consistent for AWS and Ansible sources.

Changing both to `username` to match the column name in the authentications table. This change will automatically update existing, seeded schema, when sources-api pod starts

https://projects.engineering.redhat.com/browse/TPINVTRY-533

/cc @rvsia


agrare:
@rvsia I assume this will require a change to sources-ui to use the new field?

rvsia:
> @rvsia I assume this will require a change to sources-ui to use the new field?

@agrare Yes, using API matching names would be the best way to reach consistency and best maintability.

I will prepare a UI PR tomorrow.

PR #58, opened 2019-07-25
Fix missing dependent destroy from sources

agrare:
We were missing the dependent destroy on the sources -> endpoints ->
authentications associations.  This was causing orphaned Authentication
records because the polymorphic authentication relationship couldn't do
a foreign key triggered cascade delete.


agrare:
I think long term we should move away from authentication being a polymorphic association, originally it was this way so that it could be linked to sources, hosts, vms, etc... but now that this is only in sources it really only can be tied to an endpoint, and then we could use normal FK triggered cascade destroy.

PR #59, opened 2019-07-25
Source availability_status

abellotti:
Adding support for Source availability_checker

- exposing the availability_status attribute
- rspecs to verify patch updates of the availability_status attribute

The new sources-monitor project to periodically kick off availability checks on sources and messaging to appropriate kafka queues when sources need checking. This new project  schedules tasks for running availability checks on available and unavailable sources via OpenShift CronJob ephemeral pods.
https://github.com/abellotti/sources-monitor

Which also requires the updated ruby client:
https://github.com/ManageIQ/sources-api-client-ruby/pull/5

And the receiving side getting and processing those messages:
https://github.com/ManageIQ/topological_inventory-openshift/pull/95
https://github.com/ManageIQ/topological_inventory-amazon/pull/32

And orchestrator which honors the availability_status for considering kicking of a collector for a source:
https://github.com/ManageIQ/topological_inventory-orchestrator/pull/32

The updated e2e-deploy project for building and deploying the new sources-monitor project:
https://github.com/RedHatInsights/e2e-deploy/pull/605


abellotti:
TODO:

- [ ] targeting different topic based on type

abellotti:
@miq-bot assign @agrare

config/initializers/availability_checker.rb
```diff
@@ -0,0 +1,7 @@
+# Instantiate the Source Availability Checker Singleton
+Rails.application.config.after_initialize do
+  if defined?(::Rails::Server)
+    SourceAvailabilityChecker.instance
```
agrare: Ehh idk about a thread in the API server, if we scale this up we'll get one of these per pod.  I pictured the actual scheduling of the availability check as a separate worker.  @Fryguy ?

lib/source_availability_checker.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def check_available_sources
+    check_sources = []
+
+    Source.includes(:source_type, :tenant).all.each do |source|
+      next unless source_available(source)
+
+      check_sources << source.id
+      Sources::Api::Events.send_event(
```
agrare: This isn't really an event at this point since it is going to the operations worker it is more of a queue message, I know it is just naming but the Event class was supposed to be very specific not just reusing the kafka connection.

config/initializers/availability_checker.rb
abellotti: makes sense. we need one and only one of those. 👍

abellotti: Marking all as WIP for now. One other potential issue bugging me this morning is with
https://github.com/ManageIQ/topological_inventory-orchestrator/pull/32 coming in the picture.
Let's say we resolve the one and only one instance (worker, etc), which is needed. Next, sources being unavailable, the worker sees that and sends a message to the operation worker side of things, he does his magic and sees a source up and reachable, then he sends a request to update the source's availability_status, the patch might fail because there may not be a source running to take that api call because the orchestrator didn't bring one up (availability_status being down).

abellotti: Ok, re: ^ I think we're good as the orchestrator only brings up/down collectors, so the sources project and the operations workers would always be running.

agrare: @abellotti yeah the operations workers are per type not per source :+1:

abellotti:
Ok, simplifying this a bit.  

Removing the availability checker and related threads from the sources-api.  This will now be handled externally via a worker and triggered via OpenShift in the sources-monitor project.
    
This PR now just implements the availability_status attribute, JSON and specs.

app/models/source.rb
```diff
@@ -6,6 +6,9 @@ class Source < ApplicationRecord
   has_many :application_types, :through => :applications
   has_many :endpoints, :autosave => true, :dependent => :destroy
 
+  attribute :availability_status, :string
+  validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
```
agrare: Since this is going to be an aggregated status (eventually we'd have status on the endpoints and authentications) can we have an intermediate status like mixed or partially_available

PR #62, opened 2019-08-06
Add Azure source

Ladas:
Adding azure source and ability to send authentication with {"extra": {"azure": {"tenant_id": "some id"}}}

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -122,8 +122,8 @@ class ApplicationController < ActionController::API
   end
 
   def params_for_create
-    required = api_doc_definition.required_attributes
-    body_params.permit(*api_doc_definition.all_attributes).tap { |i| i.require(required) if required }
+    # We already validate this with OpenAPI validator, that validates every request, so we shouldn't do it again here.
+    body_params.permit!
   end
 
   def safe_params_for_list
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -31,6 +31,18 @@ amazon_json_schema = {
 }
 update_or_create(SourceType, :name => "amazon", :product_name => "Amazon Web Services", :vendor => "Amazon", :schema => amazon_json_schema)
 
+azure_json_schema = {
+  :title  => "Configure Azure",
+  :fields => [
+    {:component => "text-field", :name => "role", :type => "hidden", :initialValue => "azure"},
+    {:component => "text-field", :name => "authtype", :type => "hidden", :initialValue => "access_key_secret_key"},
+    {:component => "text-field", :name => "tenant_id", :label => "Tenant ID"},
+    {:component => "text-field", :name => "username", :label => "Client ID"},
+    {:component => "text-field", :name => "password", :label => "Client Secret", :type => "password"}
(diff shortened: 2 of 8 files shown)


agrare:
@Ladas is the azure tenant_id something that would be required by any service connecting to Azure or something specific to e.g. topology collection?
Is it something that could be encoded in the endpoint?

Basically wondering do we need to add another column to the sources table

Ladas:
@agrare you need the tenant_id for any Azure API call (you also need subscription ID, but we introspect that)

So yeah we need either separate column, or jsonb for modeling extra things needed for a specific provider.

agrare:
@Ladas okay I like the jsonb column with the schema defined in the source_type record

app/controllers/application_controller.rb
```diff
@@ -122,8 +122,8 @@ def raise_event(event, payload)
   end
 
   def params_for_create
-    required = api_doc_definition.required_attributes
-    body_params.permit(*api_doc_definition.all_attributes).tap { |i| i.require(required) if required }
+    # We already validate this with OpenAPI validator, that validates every request, so we shouldn't do it again here.
+    body_params.permit!
```
Ladas: @bdunne I am disabling the validation here, since we validate a superset of this for each request, using the openapi validator. (which also validates nested objects, like the `{:extra => {:azure => {:tenant_id => "xy"}}}' that I am adding here)

db/seeds.rb
```diff
@@ -31,6 +31,18 @@ def update_or_create(model, attributes)
 }
 update_or_create(SourceType, :name => "amazon", :product_name => "Amazon Web Services", :vendor => "Amazon", :schema => amazon_json_schema)
 
+azure_json_schema = {
+  :title  => "Configure Azure",
+  :fields => [
+    {:component => "text-field", :name => "role", :type => "hidden", :initialValue => "azure"},
+    {:component => "text-field", :name => "authtype", :type => "hidden", :initialValue => "access_key_secret_key"},
+    {:component => "text-field", :name => "extra.azure.tenant_id", :label => "Tenant ID"},
```
Ladas: @martinpovolny  @rvsia is this the right format for the component? I want this value to be sent as `:extra => {:azure => {:tenant_id => "tenant_id_value"}}.to_json`

like you can see in the spec here https://github.com/ManageIQ/sources-api/pull/62/files#diff-14ea593593f9669c3585d47cacefe071R58

agrare: @martinpovolny can you check that these params are correct?

app/controllers/application_controller.rb
bdunne: Do we have a test that raises an error when a required field is missing?

Ladas: Hm, not in here. We don't have any required fields in the spec. (I test it in other repo though, so we know it works)

I was trying to add the required fields. But that won't work since we e.g. use the same model in patch query. So first we will need to do a schema refactoring, where post/put/patch will have different readOnly and required definitions,

agrare:
Lets get this in and if the params are wrong we can fix them

PR #64, opened 2019-08-09
Filter tenant and tenant_id params from create/update requests

gtanzillo:
Fixes https://projects.engineering.redhat.com/browse/TPINVTRY-416

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,8 @@ class ApplicationController < ActionController::API
   include ManageIQ::API::Common::ApplicationControllerMixins::RequestBodyValidation
   include ManageIQ::API::Common::ApplicationControllerMixins::RequestPath
 
+  BLACKLIST_PARAMS = [:tenant, :tenant_id]
+
   around_action :with_current_request
 
   rescue_from ActiveRecord::RecordNotFound do |exception|
@@ -73,7 +75,7 @@ class ApplicationController < ActionController::API
 
   def params_for_create
     # We already validate this with OpenAPI validator, that validates every request, so we shouldn't do it again here.
-    body_params.permit!
+    body_params.reject {|k, _| BLACKLIST_PARAMS.include?(k.to_sym)}.permit!
   end
 
   def safe_params_for_list
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::API
   include ManageIQ::API::Common::ApplicationControllerMixins::RequestBodyValidation
   include ManageIQ::API::Common::ApplicationControllerMixins::RequestPath
 
-  BLACKLIST_PARAMS = [ :tenant ]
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -154,6 +156,6 @@ def pagination_offset
   end
 
   def params_for_update
-    body_params.permit(*api_doc_definition.all_attributes - api_doc_definition.read_only_attributes)
+    body_params.except(*BLACKLIST_PARAMS).permit(*api_doc_definition.all_attributes - api_doc_definition.read_only_attributes)
```
agrare: I don't think this is needed, without it:

```
curl -X PATCH http://localhost:3000/api/v1.0/sources/1 -d '{"tenant": "5678"}' 2>/dev/null | python -m json.tool
{
    "errors": [
        {
            "detail": "found unpermitted parameter: :tenant",
            "status": 400
        }
    ]
}
```

and
```
curl -X PATCH http://localhost:3000/api/v1.0/sources/1 -d '{"tenant_id": "2"}' 2>/dev/null | python -m json.tool
{
    "errors": [
        {
            "detail": "properties tenant_id are not defined in #/components/schemas/Source",
            "status": 400
        }
    ]
}
```

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,8 @@ class ApplicationController < ActionController::API
   include ManageIQ::API::Common::ApplicationControllerMixins::RequestBodyValidation
   include ManageIQ::API::Common::ApplicationControllerMixins::RequestPath
 
+  BLACKLIST_PARAMS = [:tenant, :tenant_id]
```
agrare: I don't think tenant_id needs to be here either because it isn't in the openapi doc so:
```
curl -X POST http://localhost:3000/api/v1.0/sources -d '{"source_type_id": "1", "name": "OCP", "tenant_id": "5678"}' 2>/dev/null | python -m json.tool
{
    "errors": [
        {
            "detail": "properties tenant_id are not defined in #/components/schemas/Source",
            "status": 400
        }
    ]
}
```

app/controllers/application_controller.rb
gtanzillo: @agrare I see that, but do we want to return the 400 here? Was thinking we'd still strip out the `tenant` param, essentially ignoring it so that it would still succeed. What do you think?

agrare: Oh, I don't know...if it is read-only then I think trying to update it should be a bad request error.  @bdunne?

bdunne: I agree, I think we need to return a 400 here because we say [additional properties are not allowed](https://github.com/ManageIQ/sources-api/blob/42b6081347b45958898121f39197007b8b1d23ad/public/doc/openapi-3-v1.0.0.json#L1366)

gtanzillo: Thanks, Adam, Brandon, I'll change it

agrare:
Just those two rubocop issues then lgtm

PR #65, opened 2019-08-14
Add SourceTypeIcons Model and API

agrare:
Allow an Icon to be added for the different source_types

https://projects.engineering.redhat.com/browse/TPINVTRY-535


agrare:
cc @Hyperkid123 

I based this off of the service_offering_icons

agrare:
`curl http://localhost:3000/api/v1.0/source_types/1/source_type_icon` downloads the icon data

bdunne:
Would we ever need the relation to the icon record rather than just the image? Will there be other data associated with it that we will need? In topo we have `/service_offering_icons/{id}/icon_data` to get just the icon data.
How are these icons added to the database? (seed, HTTP POST?)

agrare:
> How are these icons added to the database? (seed, HTTP POST?)

I was imagining we'd import these with db seed like we do with the source_types

> Would we ever need the relation to the icon record rather than just the image? Will there be other data associated with it that we will need? In topo we have /service_offering_icons/{id}/icon_data to get just the icon data.

Yeah this is how I had it originally but wanted the has_one under a source_type, I can add the index for `/source_type_icons` to get all of the icons

bdunne:
> > How are these icons added to the database? (seed, HTTP POST?)
> 
> I was imagining we'd import these with db seed like we do with the source_types

In that case, should we serve static files instead?

agrare:
Yeah that would work

Hyperkid123:
> > > How are these icons added to the database? (seed, HTTP POST?)
> > 
> > 
> > I was imagining we'd import these with db seed like we do with the source_types
> 
> In that case, should we serve static files instead?

Serving static files would be even better. We would be able to leverage all the caching goodies that browsers provide.

Hyperkid123:
@agrare Returning the URL for a static file would work perfectly.

For the location of the file. I think it might be a good idea to have some separated storage for all static files in across the platform. There is a place for hosting static assets already but its not configured for images etc. just for some config files. I will open this discussion with platform folks and we should have something soon.

But for now i think the API pod should be enough.

terezanovotna:
Opened an issue in pf as it would be great to grab icons from Patternfly
https://github.com/patternfly/patternfly-react/issues/2795

agrare:
Thanks @terezanovotna :+1:

bdunne:
@terezanovotna Are you suggesting adding icons for AWS, Ansible Tower, Vmware, etc. to Patternfly?  That didn't fit with my understanding of Patternfly's purpose.  I was expecting to drop the icons in the public directory of this repo and seed the records.

agrare:
Alright instead of a new table I just added an `icon_url` to `source_types`, and added three example icons.

agrare:
@Hyperkid123 will this work for you?

Hyperkid123:
@agrare yes for now it will. I will get back to you once we have platform CDN.

PR #73, opened 2019-08-29
Split seeds into per-model files

agrare:
The single db/seeds.rb file was getting a little crowded, splitting it up
into per-model files should help clean things up.

~~Depends on: https://github.com/ManageIQ/sources-api/pull/65~~


db/seeds.rb
```diff
@@ -1,88 +1,2 @@
...
-                 :dependent_applications         => [],
-                 :supported_source_types         => ["amazon", "ansible_tower", "azure", "openshift"],
-                 :supported_authentication_types => {
-                   "amazon"        => ["access_key_secret_key"],
-                   "ansible_tower" => ["username_password"],
-                   "azure"         => ["username_password"],
-                   "openshift"     => ["token"]
-                 })
+# Run all seed files in db/seeds/*.rb
+Dir[File.join(Rails.root.join("db", "seeds"), "**/*.rb")].each { |f| load f }
```
bdunne: Does `Dir[Rails.root.join("db", "seeds", "**", "*.rb")].each { |f| load f }` work?

bdunne: Should we `require` rather than `load`?

agrare: > Does Dir[Rails.root.join("db", "seeds", "**", "*.rb")].each { |f| load f } work?

Yes

agrare: > Should we require rather than load?

:+1: Done

agrare:
@bdunne I addressed the rubocop issues in separate commits since originally this was a simple move to a new file

PR #74, opened 2019-09-02
Rename fields in DDF form schemas

rvsia:
**Description**

Renames form fields in DDF to allow better handling in UI.

@Ladas Is the  `authentication.extra.azure.tenant_id` the right endpoint (authentication) ?

Merge together with Sources-UI (https://github.com/ManageIQ/sources-ui/pull/94), and after frontend-components-sources (https://github.com/RedHatInsights/frontend-components/pull/132)

### DIFF
--- a/db/seeds/source_types.rb
+++ b/db/seeds/source_types.rb
@@ -10,12 +10,12 @@ end
 openshift_json_schema = {
   :title  => "Configure OpenShift",
   :fields => [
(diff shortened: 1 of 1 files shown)


Ladas:
@rvsia yes, it should be correct

rvsia:
Merge before/together with https://github.com/ManageIQ/sources-ui/pull/94

PR #75, opened 2019-09-04
Update openapi-generator-cli version to 4.1.1

agrare:
The 4.1.1 version of openapi-generator-cli was released and we're a few
versions behind.

### DIFF
--- a/lib/tasks/client_generate.rake
+++ b/lib/tasks/client_generate.rake
@@ -6,7 +6,7 @@ class ClientGenerator
   require 'json'
   require 'uri'
 
-  VERSION = "3.3.4".freeze
+  VERSION = "4.1.1".freeze
   SOURCE_URL = "http://central.maven.org/maven2/org/openapitools/openapi-generator-cli".freeze
 
   def msg(message)


slemrmartin:
@agrare are there some changes in generated clients we should be aware?

agrare:
@slemrmartin I'm going to push a PR to sources-api-client-ruby shortly, but from a high level it looks like the biggest change is that you can specify e.g. `opts[:query_params]` or `opts[:header_params]` with the method calls e.g.:
```
     def create_authentication_with_http_info(authentication, opts = {})
       if @api_client.config.debugging
         @api_client.config.logger.debug 'Calling API: DefaultApi.create_authentication ...'
@@ -101,28 +110,37 @@ module SourcesApiClient
       local_var_path = '/authentications'
 
       # query parameters
-      query_params = {}
+      query_params = opts[:query_params] || {}
```

agrare:
@slemrmartin https://github.com/ManageIQ/sources-api-client-ruby/pull/8

slemrmartin:
looks good, also `return_type` can be customized

agrare:
Yeah that should help us on the ingress_api where we had to manually update it

PR #78, opened 2019-09-13
Add URL validator to DDF schemas

rvsia:
**Description**

- Adds an URL validator for URL fields

Accepts all kinds of address (ipv4, ipv6, with/out port, etc.), the only part which is required is protocol identifier. 

@gtanzillo Which protocols are we going to support? Right now, it is set to accept only `http`,`https`, but anything else could be added. Or we could make the protocol optional too.

### DIFF
--- a/db/seeds/source_types.rb
+++ b/db/seeds/source_types.rb
@@ -12,7 +12,7 @@ openshift_json_schema = {
   :fields => [
     {:component => "text-field", :name => "endpoint.role", :type => "hidden", :initialValue => "kubernetes"},
     {:component => "text-field", :name => "authentication.authtype", :type => "hidden", :initialValue => "token"},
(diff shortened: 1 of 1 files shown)


gtanzillo:
@rvsia 
>@gtanzillo Which protocols are we going to support? Right now, it is set to accept only http,https, but anything else could be added. Or we could make the protocol optional too.

I think that's all we need to support @agrare?

agrare:
I'm not sure what protocol we'd need for receptor-over-http but I don't think the receptor team knows either so we can cross that bridge when we get there.  http+https is good for now

rvsia:
Good to hear it, so this PR should be fine then.

rvsia:
Closes TPINVTRY-351

PR #86, opened 2019-09-20
Fix typo in application_types.db

rvsia:
In application_types it is `ansible_tower`, in source types it is `ansible-tower`.

I changed it to `ansible_tower` in source-types. I am not sure what naming system are we going to use, but it makes sense to use the name with `_` as it is a standard Ruby syntax.

Do I need to write any migration? 

@gtanzillo @agrare

### DIFF
--- a/db/seeds/application_types.rb
+++ b/db/seeds/application_types.rb
@@ -11,8 +11,8 @@ update_or_create(
   :name                           => "/insights/platform/catalog",
   :display_name                   => "Catalog",
   :dependent_applications         => ["/insights/platform/topological-inventory"],
-  :supported_source_types         => ["ansible_tower"],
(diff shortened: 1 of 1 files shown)


rvsia:
The test failure does not seem to be related with this PR.

agrare:
:+1: good catch @rvsia

agrare:
I'll work on the spec failure, and no you don't need a migration the seeds are updated automatically

rvsia:
@agrare Thanks. I removed the WIP label from this. :+1:

agrare:
@rvsia actually the orchestrator is looking for `"ansible-tower"` today, can you update application_types to use `"ansible-tower"`?

rvsia:
@agrare done! :+1:

PR #89, opened 2019-09-23
Now empty bodies are allowed by the API

agrare:
ManageIQ/manageiq-api-common#93 allows empty bodies so these tests no
longer match the api-common expectations.


agrare:
@gtanzillo this gets sources-api green again

PR #91, opened 2019-09-24
Add return type to getDocumentation /openapi.json

agrare:
There was no return type defined for getDocumentation so generated
clients were not picking this endpoint up.

TPINVTRY-515


agrare:
Client update https://github.com/ManageIQ/sources-api-client-ruby/pull/11

PR #95, opened 2019-10-01
Add 404 return to OpenAPI spec for collections

gtanzillo:
https://projects.engineering.redhat.com/browse/TPINVTRY-232

### DIFF
--- a/public/doc/openapi-3-v1.0.0.json
+++ b/public/doc/openapi-3-v1.0.0.json
@@ -46,16 +46,6 @@
                 }
               }
             }
-          },
-          "404": {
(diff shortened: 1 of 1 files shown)


agrare:
@gtanzillo we should probably update the open_api::generator https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/open_api/generator.rb#L172-L178 to add 404 if the primary_collection is non-nil

agrare:
Just need @bdunne to release api-common to include your 404 updates

bdunne:
v2.0.0 was released today

PR #97, opened 2019-10-04
Split schemas

rvsia:
closes https://github.com/ManageIQ/sources-api/issues/88

@agrare @gtanzillo 

**MERGE TOGETHER WITH UI PRs**: https://github.com/ManageIQ/sources-ui/pull/111

**Format**

```rb
schema = {
  :authentication => [{
    :type             => 'username_password', # same as in fields - authtype
    :name             => 'Token', # Will be shown in the authentication selection step
    :fields           => [
      # DDF Fields, will be shown in the authentication selection step
      {:component => "text-field", :name => "authentication.authtype", :type => "hidden", :initialValue => "username_password"}
    ],
    :additional_steps => [
      # Steps according to DDF
      # Don't use `stepKey` in the first step
      # Don't use `nextStep` in the last step
      # => These attributes will be appended automatically
      # Additional steps are shown after the authentication_type selection step
    ]
  }],
  :endpoint       => {
    :title  => 'Title', # If the endpoint will be on a separate step
    :hidden => true, # If true, the endpoint step will be hidden, use if all components have `:hideField => true`
    :fields => [
      # DDF Fields
      {:component => "checkbox", :name => "endpoint.verify_ssl", :label => "Verify SSL"}
    ]
  }
}

```

### DIFF
--- a/db/seeds/source_types.rb
+++ b/db/seeds/source_types.rb
@@ -8,15 +8,23 @@ def update_or_create(attributes)
 end
 
 openshift_json_schema = {
-  :title  => "Configure OpenShift",
-  :fields => [
-    {:component => "text-field", :name => "endpoint.role", :type => "hidden", :initialValue => "kubernetes"},
-    {:component => "text-field", :name => "authentication.authtype", :type => "hidden", :initialValue => "token"},
-    {:component => "text-field", :name => "url", :label => "URL", :validate => [{:type => "url-validator"}]},
-    {:component => "checkbox", :name => "endpoint.verify_ssl", :label => "Verify SSL"},
-    {:component => "text-field", :name => "endpoint.certificate_authority", :label => "Certificate Authority", :condition => {:when => "endpoint.verify_ssl", :is => true}},
-    {:component => "text-field", :name => "authentication.password", :label => "Token", :type => "password"}
-  ]
+  :authentication => [{
+    :type   => 'token',
+    :name   => "Token",
+    :fields => [
+      {:component => "text-field", :name => "authentication.authtype", :type => "hidden", :initialValue => "token"},
+      {:component => "text-field", :name => "authentication.password", :label => "Token", :type => "password"}
+    ]
+  }],
+  :endpoint       => {
+    :title  => "Configure OpenShift endpoint",
+    :fields => [
+      {:component => "text-field", :name => "endpoint.role", :type => "hidden", :initialValue => "kubernetes"},
+      {:component => "text-field", :name => "url", :label => "URL", :validate => [{:type => "url-validator"}]},
+      {:component => "checkbox", :name => "endpoint.verify_ssl", :label => "Verify SSL"},
+      {:component => "text-field", :name => "endpoint.certificate_authority", :label => "Certificate Authority", :condition => {:when => "endpoint.verify_ssl", :is => true}}
+    ]
+  }
 }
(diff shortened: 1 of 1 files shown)


rvsia:
@gtanzillo @agrare This is the format I used to test https://github.com/RedHatInsights/frontend-components/pull/190 Please, review this now to get all possible problems away (however wait with merging until I prepare a Sources-UI PR.)

db/seeds/source_types.rb
```diff
@@ -8,15 +8,29 @@ def update_or_create(attributes)
...
+        :fields => [
+          {:component => "text-field", :name => "authentication.authtype", :type => "hidden", :initialValue => "token"},
+          {:component => "text-field", :name => "authentication.password", :label => "Token", :type => "password"}
+        ]
+      }
+    }
+  },
+  :endpoint       => {
+    :title  => "Configure OpenShift endpoint",
+    :fields => [
```
slemrmartin: is it possible to have similar path for both models? I mean specifically `fields`:
**authentication**: 
* `/<type>/:schema/:fields`

**endpoint**:
* `/fields`

instead -> `/:schema/:fields` for both

rvsia: Actually, I am removing the `:schema` from authentication right now, as it's not useful.

rvsia: I have put the format into the description, what do you think?

slemrmartin: nice :+1: 
I see 2nd option in authentication, instead of 
```
:name_of_the_authentication_type => {}, 
:name_of_the_authentication_type => {},
```
have an array: 
`[ { :type => ... }, { :type => ... } ]`
I think it's more formal description, but your way is probably easier to implement.
@agrare what do you think?

agrare: @slemrmartin yeah it does look a little strange having seemingly an arbitrary key there.  I assume @rvsia is getting the key to use from the authentication_type schema but still I'd agree that an array of types looks better

rvsia: Yeah, with keys it is slightly simpler to use, but I have no problem with changing it.

db/seeds/source_types.rb
```diff
@@ -8,15 +8,25 @@ def update_or_create(attributes)
...
-    {:component => "text-field", :name => "endpoint.role", :type => "hidden", :initialValue => "kubernetes"},
-    {:component => "text-field", :name => "authentication.authtype", :type => "hidden", :initialValue => "token"},
-    {:component => "text-field", :name => "url", :label => "URL", :validate => [{:type => "url-validator"}]},
-    {:component => "checkbox", :name => "endpoint.verify_ssl", :label => "Verify SSL"},
-    {:component => "text-field", :name => "endpoint.certificate_authority", :label => "Certificate Authority", :condition => {:when => "endpoint.verify_ssl", :is => true}},
-    {:component => "text-field", :name => "authentication.password", :label => "Token", :type => "password"}
-  ]
+  :authentication => [{
+    :type   => 'token',
+    :meta   => {
```
agrare: What else might be in this `:meta` section?  I'm with @slemrmartin any reason not just to move `:name` (or `:label` like the `:fields`)

rvsia: I initially thought that there would be some description.

I think `:name` is good choice.

PR #98, opened 2019-10-07
Update manageiq-api-common dependency

bdunne:
Fixes TPINVTRY-639

Also pulls in v1.0.0 changes:
### Breaking Changes
- Namespace dynamically defined GraphQL API version modules under ManageIQ::API::Common::GraphQL::Api #105

### Added
- New entitlements for Ansible and Migrations #113 and use them #115
- Forwarding of persona headers #109

### Changed
- Extract GraphQLRequest to a schema object rather than inline #110
- Update to openapi_parser v0.5.0 #108

### Fixed
- Fixed errors around already defined constants in GraphQL #112
- Allow GraphQL generator to support paths with IDs like /primary/{primary_id}/ #107
- Constant problems related to GraphQL namespace #105

### DIFF
--- a/public/doc/openapi-3-v1.0.0.json
+++ b/public/doc/openapi-3-v1.0.0.json
@@ -525,27 +525,7 @@
           "content": {
             "application/json": {
               "schema": {
-                "type": "object",
-                "properties": {
-                  "query": {
-                    "type": "string",
-                    "description": "The GraphQL query",
-                    "default": "{}"
-                  },
-                  "operationName": {
-                    "type": "string",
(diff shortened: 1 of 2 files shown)


public/doc/openapi-3-v1.0.0.json
```diff
@@ -525,27 +525,6 @@
           "content": {
             "application/json": {
               "schema": {
-                "type": "object",
```
bdunne: @abellotti Looks like we're missing the `$ref` to the schema object

PR #102, opened 2019-10-10
Add imported column to sources table

lpichler:
TPINVTRY-632
after UX discussion  we decided to have read-only column here, see comments and mockup in the ticket.

source api client https://github.com/ManageIQ/sources-api-client-ruby/pull/12

@miq-bot assign @agrare

### DIFF
--- a/db/migrate/20191011124345_add_imported_to_sources.rb
+++ b/db/migrate/20191011124345_add_imported_to_sources.rb
@@ -1,0 +1,5 @@
+class AddImportedToSources < ActiveRecord::Migration[5.2]
+  def change
+    add_column :sources, :imported, :string
+  end
+end
--- a/public/doc/openapi-3-v1.0.0.json
+++ b/public/doc/openapi-3-v1.0.0.json
@@ -1335,6 +1335,9 @@
           "id": {
             "$ref": "#/components/schemas/ID"
           },
+          "imported": {
+            "type": "string"
+          },
           "name": {
             "example": "Sample Provider",
             "title": "Name",


agrare:
Hm I'm not sure that "read-only" is what I'd call this.  You might still want to set different applications to make use of the CFME data which I wouldn't expect to do for a "read-only" source.

slemrmartin:
Also it should be possible to delete from UI (when customer wants to stop sync from CF), so I'm not sure about "read-only" flag in both db and UI too.   
And import through sync-processor should not only check existence, but it should update name if changed in CF, so from API's point of view, it's editable so db can use "imported" flag regardless how it is called in UI

lpichler:
@agrare, as @slemrmartin said, are you ok with `imported` ? If so I will change related PRs

thanks!

agrare:
If we just want to show that a source was imported then I'm good with that.

If we care about what it was imported from then I think a label would be better, e.g. we could have a source imported from CFME, from Ansible Analytics, or from Openshift Online.

terezanovotna:
This clarifies the use case. We definitely want to tell the user that this source is somehow different than other sources as it really cannot be edited. 

I suggest using "imported" label - sounds like a good wording? and then on a pop up give more detail where this comes from - CFME, Ansible Analytics etc.

slemrmartin:
then label can be "imported": string. When blank, then it's editable in UI, when non-blank then what @terezanovotna wrote

agrare:
Yup, and the label val could be what it is imported from

slemrmartin:
cc @rvsia

lpichler:
done 👍  (also related PRs are updated )

agrare:
I was picturing an actual tag but this works too, side note we don't actually have tags/taggings on sources at all (?!) We should get on that

PR #103, opened 2019-10-10
Fix Azure authentication_type

agrare:
The Azure authentication_type was an access_key_secret_key type even
though the application_type specified username_password and azure also
includes a tenant_id field.

Introduce a new authentication_type to track the azure credential type.

Fixes https://github.com/ManageIQ/sources-api/issues/101


agrare:
cc @rvsia

agrare:
cc @dccurtis I don't know if you're referencing this auth_type specifically or just using any azure authentication

PR #107, opened 2019-10-17
Add arn auth_type for aws+cm

rvsia:
**Description**

- Adds ARN DDF schema for ARN for AWS/CM.
    - basic structure and input fields
    - Tags step is skipped until we get exact steps

`description` component is a component, where the content is added on the UI side, because it's a complex task to write these components into the schema without using Javascript syntax. (It uses bullet lists, copy clipboard components, etc.) However, if we decide that we want to store everything in the DB, it can be done, but it will take some time to come up with a good solution. (We could probably used some `md` parser... Client Side has advantages: it's simple and it can be easily localized.) 

~~_Waiting for completion of the UI implementation._~~

Work-in-progress: https://github.com/RedHatInsights/frontend-components/pull/212

### DIFF
--- a/db/seeds/source_types.rb
+++ b/db/seeds/source_types.rb
@@ -44,6 +44,36 @@ amazon_json_schema = {
       {:component => "text-field", :name => "authentication.username", :label => "Access Key"},
       {:component => "text-field", :name => "authentication.password", :label => "Secret Key", :type => "password"}
     ]
+  }, {
+    :type   => 'arn',
+    :name   => 'ARN',
+    :fields => [{
+      :component    => 'text-field',
+      :name         => 'authentication.authtype',
+      :hideField    => true,
+      :initialValue => 'arn'
+    }, {
+      :name       => 'billing_source.bucket',
+      :stepKey    => 'amazon-arn-additional-step',
+      :component  => 'text-field',
+      :label      => 'S3 bucket name',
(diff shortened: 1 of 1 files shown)


agrare:
@rvsia I'd rather not store _everything_ in the database, just what is needed for accepting input and doing validations.  All of the doc text, etc... sounds like it'd be better done in the frontend.

agrare:
Hakiri should be fixed now

agrare:
@rvsia do we need the `:additional_steps` defined here if they don't actually take any input?  Or could they be defined only in the UI as well?

rvsia:
> @rvsia do we need the `:additional_steps` defined here if they don't actually take any input? Or could they be defined only in the UI as well?

In the last step, there is the ARN input defined. I think we could add a `step` key to the objects in fields, and then inject these fields into the right step in UI, however, we would lose information about the structure of the form in the API.

agrare:
Right so could we only define the ARN input schema and leave the rest to the UI since it is really just display anyway without any API interaction?

rvsia:
@agrare will do! 👍

rvsia:
@agrare So I updated the schema: it does not contain `additionalFields` anymore. The UI is now responsible for the structure, the API can only tell the step where the field is. Fields without defined `stepKey` will be shown in the auth selection page.

PR #112, opened 2019-10-21
Add test for Authentication patch with extra attributes

agrare:
Depends on: ~~https://github.com/ManageIQ/manageiq-api-common/pull/129~~, https://github.com/ManageIQ/sources-api/pull/115


agrare:
@bdunne please review

PR #116, opened 2019-10-28
Add AvailabilityStatus to Applications, Authentications, and Endpoints

agrare:
Add AvailabilityStatus to Applications, Authentications, and Endpoints.

Extend the granularity of availability_status by allowing more resources
to contain availability status and an error message.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -2,4 +2,7 @@ class Application < ApplicationRecord
   include TenancyConcern
   belongs_to :source
   belongs_to :application_type
+
+  attribute :availability_status, :string
+  validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
 end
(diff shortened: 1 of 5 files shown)


app/models/application.rb
```diff
@@ -2,4 +2,7 @@ class Application < ApplicationRecord
   include TenancyConcern
   belongs_to :source
   belongs_to :application_type
+
+  attribute :availability_status, :string
```
bdunne: Any reason we need to set the `cast_type` on these strings?

agrare: I copied what was already in [Source](https://github.com/ManageIQ/sources-api/blob/master/app/models/source.rb#L9)
We can change it but would like to keep it consistent across the board

slemrmartin:
@agrare what is the use case of AvailabilityStatus for Application? Either the Source is unavailable or Application is unavailable, but I cannot imagine why Application will be unavailable for one Source

agrare:
@slemrmartin `Application` is the link between a `Source` and an `ApplicationType` like `catalog` or `cost-management`.

`availability_status` on an application can indicate for example that the ARN credentials for cost-management are fine, but the S3 bucket (app specific) is missing.

PR #117, opened 2019-10-31
Adding sources_uid to sources table to support Cost Management OCP integration

dccurtis:
For more detailed discussion on this see https://github.com/project-koku/koku/pull/1319.

Test results are posted there as well.  Open to any name or placement changes that you want to see.  Thanks for helping out on this!


slemrmartin:
@dccurtis I'm note sure how far I understand that discussion, but I don't like column `source_uid` when the source have its own `uid`. It's confusing. What about cluster_uid?
And it seems this is cost-management svc specific, so saving this value in `applications` table which defines Source <-> cost_management association seems better for me

agrare:
> but I don't like column source_uid when the source have its own uid. It's confusing. What about cluster_uid?

Yeah it is a little confusing (I suck at naming things) but I'd rather not cluster because this can be used by sources other than OCP.  Maybe source_ref is better, we usually use source_ref as the "native unique reference".

> And it seems this is cost-management svc specific

Actually I think this can be used elsewhere, when syncing cfme sources we overloaded the `:uid` column but I think this would be a better way to do it (e.g.: we ended up having issues with two tenants trying to sync the same CFME source because of the unique constraint)

slemrmartin:
yeah, `source_ref` is better (although other models are using `source_id` as sources-db reference and `source_ref` as remote reference, it's at least consistent)

dccurtis:
`source_ref` sounds good to me.  As just a user of the API the multiple concepts of an id did trip me up once, and I wrote it!

app/models/source.rb
```diff
@@ -15,6 +15,7 @@ class Source < ApplicationRecord
            :to => :default_endpoint, :allow_nil => true
   validates :name, :presence => true, :allow_blank => false,
             :uniqueness => { :scope => :tenant_id }
+  attribute :source_ref, :string
```
agrare: This isn't necessary unless you're trying to change the way the attribute is serialized to SQL, rails defines these attributes automatically from the db columns

public/doc/openapi-3-v1.0.0.json
```diff
@@ -1531,6 +1531,9 @@
           "imported": {
             "type": "string"
           },
+          "source_ref": {
+            "type": "string"
+          },
```
agrare: Can you use the openapi generator task? `bin/rake openapi:generate`

The content is correct but the order is out of place so it will look like more changes the next time we run the generator for something else

agrare:
Just those two minor issues then this LGTM

dccurtis:
Thanks for walking me through that 👍

agrare:
Thanks @dccurtis !

PR #128, opened 2019-11-22
Adding the POST /api/sources/:source_id/check_availability action

abellotti:
Adding the POST `/api/sources/:source_id/check_availability` action
- Action triggers an availability check on the Source[:source_id] specified
- Returns a 202 with empty {} body for now.

### DIFF
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -15,6 +15,23 @@ module Api
 
         render :json => source, :status => :created, :location => instance_link(source)
       end
+
+      def check_availability
+        source = Source.find(params[:source_id])
+
+        Sources::Api::Messaging.client.publish_topic(
+          :service => "platform.topological-inventory.operations-#{source.source_type.name}",
+          :event   => "Source.availability_check",
+          :payload => {
+            :params => {
+              :source_id       => source.id,
+              :external_tenant => source.tenant.external_tenant
+            }
+          }
+        )
+
+        render :json => source, :status => :accepted
+      end
     end
   end
 end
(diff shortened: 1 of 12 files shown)


abellotti:
marking as wip, needs some more testing and specs.

app/controllers/api/v1/sources_controller.rb
```diff
@@ -15,6 +15,23 @@ def create
...
+        source = Source.find(params[:source_id])
+
+        Sources::Api::Messaging.client.publish_topic(
+          :service => "platform.topological-inventory.operations-#{source.source_type.name}",
+          :event   => "Source.availability_check",
+          :payload => {
+            :params => {
+              :source_id       => source.id.to_s,
+              :external_tenant => source.tenant.external_tenant
+            }
```
agrare: We should create a task so the requester can track the progress of the availability check, pass the task ID in as a param, and return the ID to the caller

agrare: Scratch that, we're going to hold off on the UI blocking on the availability check

abellotti:
moved back to wip, not sure we want to return the source, if future enhancements may be to include tasks,  thinking of just the 202/no_content for this.  @agrare ?

agrare:
@abellotti yeah that sounds good to me

lib/sources/api/messaging.rb
```diff
@@ -0,0 +1,16 @@
+module Sources
+  module Api
+    module Messaging
+      def self.client
+        require "manageiq-messaging"
+
+        @client ||= ManageIQ::Messaging::Client.open(
```
agrare: Can you update `Sources::Api::Events` to use this messaging client?

PR #132, opened 2019-12-04
Add an availability check URL for applications

agrare:
Add a URL for applications to register to allow for per-app availability
checking of sources.

This will be invoked by the normal availabililty check (scheduled and
on-demand) for all enabled applications for a source.

Applications will receive a POST action to check the availability of all
related resources for the specified source by ID, then the application
will update the source via the API to set the availability_status.

### DIFF
--- a/db/migrate/20191204145726_add_availability_check_url_to_application_types.rb
+++ b/db/migrate/20191204145726_add_availability_check_url_to_application_types.rb
@@ -1,0 +1,5 @@
+class AddAvailabilityCheckUrlToApplicationTypes < ActiveRecord::Migration[5.2]
+  def change
+    add_column :application_types, :availability_check_url, :string
+  end
+end


abellotti:
Hi @agrare this is fine.

How is this to be updated ? PATCH /application_types/{id} for apps to update their url or
something else, maybe a configmap loading in this info since this is more of a system
configuration setting.

agrare:
@abellotti I was picturing a PR to update the application_type seed https://github.com/RedHatInsights/sources-api/blob/master/db/seeds/application_types.rb

abellotti:
That works 👍 , though that means having to redeploy sources anytime another app type updates their app url, there may be other options.

gtanzillo:
I wouldn't expect the URL to change very often. I would expect that to be tied to a change on the application side where there was a need to change the URL for some reason.

agrare:
Yeah, this is the same process for everything that we seed.

PR #134, opened 2019-12-04
Adding support for application availability checks

abellotti:
Now triggering application availability checks with the source check_availability action.

### DIFF
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -45,20 +45,22 @@ module Api
           url = app_type.availability_check_url
           next if url.blank?
 
-          headers = {
-            "Content-Type"  => "application/json",
-            "x-rh-identity" => Base64.strict_encode64({'identity' => { 'account_number' => source.tenant.external_tenant }}.to_json)
-          }
+          begin
+            headers = {
+              "Content-Type"  => "application/json",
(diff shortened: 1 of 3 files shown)


abellotti:
Marking as WIP because:

- [ ] Depends on https://github.com/RedHatInsights/sources-api/pull/128
- [ ] needs to be tested
- [ ] needs specs
- [ ] need to review post signature to the micro-services

app/controllers/api/v1/sources_controller.rb
```diff
@@ -15,6 +18,49 @@ def create
...
+
+          headers = {
+            "Content-Type"  => "application/json",
+            "x-rh-identity" => Base64.strict_encode64({'identity' => { 'account_number' => source.tenant.external_tenant }}.to_json)
+          }
+
+          uri = URI.parse(url)
+          net_http = Net::HTTP.new(uri.host, uri.port)
+          request  = Net::HTTP::Post.new(uri.request_uri, headers)
+          request.body = { "source_id" => source.id.to_s }.to_json
```
abellotti: /cc @agrare just initial prototype here, we need to design/agree on signature/contract with the apps.

agrare: :+1: this is what we discussed with @dccurtis

abellotti:
Ok, added specs, can you lend some 👀 @agrare ?

PR #135, opened 2019-12-09
Revert "Updated Sources::Api::Events to drop its messaging_client and use"

agrare:
This reverts commit 59a228bcb257f585608fe87b6f2a74b8a9613cff.

We started seeing issues with kafka events, it is possible that sharing this client connection is causing issues with writing to two different kafka topics


agrare:
@gtanzillo can you take a look?  Hoping that this resolves our kafka issues.

Travis is green, not sure why it says it is pending here

PR #136, opened 2019-12-11
Add Cost Management Azure fields to source_types

boaz0:
related to https://github.com/RedHatInsights/sources-ui/issues/128

Adding "Resource group", "Storage account" and "Subscription ID" fields.

//cc @nlcwong @dccurtis @rvsia

### DIFF
--- a/db/seeds/source_types.rb
+++ b/db/seeds/source_types.rb
@@ -99,6 +99,9 @@ azure_json_schema = {
     :name   => "Tenant ID, Client ID, Client Secret",
     :fields => [
       {:component => "text-field", :name => "authentication.authtype", :hideField => true, :initializeOnMount => true, :initialValue => "tenant_id_client_id_client_secret"},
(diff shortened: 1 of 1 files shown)


db/seeds/source_types.rb
```diff
@@ -99,6 +99,9 @@ def update_or_create(attributes)
     :name   => "Tenant ID, Client ID, Client Secret",
     :fields => [
       {:component => "text-field", :name => "authentication.authtype", :hideField => true, :initializeOnMount => true, :initialValue => "tenant_id_client_id_client_secret"},
+      {:component => "text-field", :name => "billing_source.data_source.storage_account", :label => "Storage account name"},
```
rvsia: Please, provide here and below some `stepKey`, I understand that you cannot provide the correct `stepKey` because fields are appended to the end of the step, but you should provide at least some key to not showing this fields in other applications. You can do something like `:stepKey => 'cost-management-specific'`

boaz0: You're right. I forgot to add them there. Thanks!

PR #137, opened 2019-12-12
Expose Satellite as source type in UI

rvsia:
@gtanzillo 

This PR exposes Satellite source type in UI.

Not sure which values we need to collect, which are required, validators, etc.


db/seeds/source_types.rb
```diff
@@ -145,4 +145,39 @@ def update_or_create(attributes)
...
+        {
+          :component         => "text-field",
+          :name              => "authentication.authtype",
+          :hideField         => true,
+          :initializeOnMount => true,
+          :initialValue      => "receptor_node"
+        },
+        {
+          :component  => "text-field",
+          :name       => "authentication.username",
```
gtanzillo: I'm thinking that this is not the userid but instead is an id the references the Satellite instance in the on prem env. I'll check on that.

agrare: If the satellite ID is the unique identifier for the satellite we have `source.source_ref` for that

gtanzillo: Thanks Adam, Richard and I are discussing it now. I'll find out for sure and post here

gtanzillo: @rvsia @agrare I've confirmed that the Satellite ID that we need to store with the sources is a GUID that is used to identify an instance of a Satellite running on prem. So `source.source_ref` would be the place to save it.

PR #138, opened 2019-12-13
Add an internal API to update the application_types

agrare:
Allow applications to update their application_type record over the
internal API

### DIFF
--- a/app/controllers/internal/v1/application_types_controller.rb
+++ b/app/controllers/internal/v1/application_types_controller.rb
@@ -1,0 +1,11 @@
+module Internal
+  module V1
+    class ApplicationTypesController < ::ApplicationController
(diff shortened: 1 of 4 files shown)


agrare:
@abellotti @bdunne without the PATCH in the openapi spec the internal API update fails, but with it the specs fail on the missing route in the main API.

Fryguy:
Note that if application type is externally modifiable, then you're going to need to change the way seed works (unless seed is *literally* first time ever)

agrare:
Yeah we can't seed the url because the hostnames will be different based on what environment we're on (CI/QA/PROD)

agrare:
Closing this in favor of pulling the hostname out of a config map.

PR #139, opened 2019-12-13
Move ApplicationTypes seeding to use a yml file

Fryguy:
I think this should be an easier format for other developers to consume as yaml is not Ruby specific.

@agrare Please review.


app/models/application_type.rb
```diff
@@ -1,4 +1,22 @@
...
+
+      seeds.each do |name, attributes|
+        if (r = records.delete(name))
+          r.update!(attributes)
+        else
+          create!(attributes.merge(:name => name))
+        end
+      end
+
+      records.values.map(&:destroy)
```
Fryguy: Note I added destroy here for consistency, so if an entry is removed from the file, it will be destroyed...I can remove this if you'd prefer, but I didn't see the harm in adding it for consistency.

agrare:
@Fryguy I know this is only for one model right now, but know that you're going to tackle source_types next.  What about moving that to a seeding mixin since if we're just straight loading from yaml files there won't be any model specific code

Fryguy:
I was *just* doing that...so yeah, I'll get it into this PR.

Fryguy:
@agrare Updated.

PR #141, opened 2019-12-13
[WIP] Updated sources-api to leverage environment for availability_check_urls

abellotti:
Updated sources-api so that we don't store the availability_check_url in the application type, but leverage it from the environment.

this way applications can define their own for the different staged environments without needing to update the database.

### DIFF
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -42,7 +42,8 @@ module Api
 
       def check_application_availability(source)
         source.application_types.each do |app_type|
-          url = app_type.availability_check_url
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/sources_controller.rb
```diff
@@ -42,7 +42,8 @@ def check_availability
 
       def check_application_availability(source)
         source.application_types.each do |app_type|
-          url = app_type.availability_check_url
+          app_env_prefix = app_type.name.split('/').last.upcase.tr('-', '_')
```
abellotti: Hi @agrare, code driven based on name, let me know if you prefer a different method.

agrare: :+1:

PR #142, opened 2019-12-16
Fix yml for Amazon ARN authentication

rvsia:
This typo caused not-rendering of these inputs in UI

cc @dccurtis


agrare:
@rvsia are the application_types okay?

rvsia:
@agrare they seems ok, I haven’t noticed anything wrong

agrare:
:+1: thanks

PR #143, opened 2019-12-17
Ensure that only one endpoint is "default"

agrare:
Previously it was possible to have multiple "default" endpoints.  This
can cause issues when code assumes that only one endpoint is default.


agrare:
cc @bdunne

PR #145, opened 2019-12-20
Only allow org-admin to be able to use the sources-api.

abellotti:
Only allow org-admin to be able to use the sources-api - All others get a 403.

For development environments, the following can be set to skip the check:

```
BYPASS_RBAC=true
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -43,10 +43,8 @@ class ApplicationController < ActionController::API
   def with_current_request
     Insights::API::Common::Request.with_request(request) do |current|
       begin
-        if Tenant.tenancy_enabled? && Insights::API::Common::RBAC::Access.enabled?
-          raise RbacError unless current.user.org_admin? || [:get, :head, :options].include?(request.request_method_symbol)
-        end
         if Tenant.tenancy_enabled? && current.required_auth?
+          raise RbacError if Insights::API::Common::RBAC::Access.enabled? && !current.user.org_admin? && !request_is_readonly
           raise Insights::API::Common::EntitlementError unless request_is_entitled?(current.entitlement)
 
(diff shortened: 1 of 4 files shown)


abellotti:
From what I saw, the following projects access the sources-api and may need updating to include the is_org_admin if not already there.

 - [ ] sources-monitor - https://github.com/RedHatInsights/sources-monitor
 - [ ] topological-inventory-api - https://github.com/RedHatInsights/topological_inventory-api
 - [ ] catalog-api - https://github.com/RedHatInsights/catalog-api
 - [ ] approval-api - https://github.com/RedHatInsights/approval-api
 - [ ] topological_inventory-openshift - https://github.com/RedHatInsights/topological_inventory-openshift
 - [ ] topological_inventory-amazon - https://github.com/RedHatInsights/topological_inventory-amazon
 - [ ] topological_inventory-ansible_tower - https://github.com/RedHatInsights/topological_inventory-ansible_tower
 - [ ] topological_inventory-providers-common - https://github.com/RedHatInsights/topological_inventory-providers-common
 - [ ] topological_inventory-orchestrator - https://github.com/RedHatInsights/topological_inventory-orchestrator
 - [ ] topological_inventory-collector-openshift - https://github.com/RedHatInsights/topological_inventory-collector-openshift

app/controllers/application_controller.rb
```diff
@@ -41,6 +43,9 @@ class ApplicationController < ActionController::API
   def with_current_request
     Insights::API::Common::Request.with_request(request) do |current|
       begin
+        if Tenant.tenancy_enabled? && Insights::API::Common::RBAC::Access.enabled?
+          raise RbacError unless current.user.org_admin?
+        end
         if Tenant.tenancy_enabled? && current.required_auth?
```
gtanzillo: Can we just collapse this check for `Tenant.tenancy_enabled?` here and above into a single check?

PR #147, opened 2020-01-07
Drop to a single puma thread and log before and after publishing to kafka

agrare:
To help investigate lockups in the sources-api drop to a single puma thread and log before and after publishing to kafka topics including the name of the topic.
This should help us identify if the source of the issue is concurrency related or using one kafka client to publish to multiple topics.

### DIFF
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -21,9 +21,11 @@ module Api
 
       def check_availability
         source = Source.find(params[:source_id])
+        topic  = "platform.topological-inventory.operations-#{source.source_type.name}"
 
+        logger.debug("publishing message to #{topic}...")
         Sources::Api::Messaging.client.publish_topic(
-          :service => "platform.topological-inventory.operations-#{source.source_type.name}",
+          :service => topic,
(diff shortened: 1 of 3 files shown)


bdunne:
Will this break metrics collection and our auto-scaling?  If the single thread is busy handling a long-running request, I don't think we'll be able to collect metrics as we currently do from the webserver and the scaling logic is dependant on the metrics gathered.

agrare:
I thought metrics were pushed from another thread (@carbonin ?), but I can add a thread_id to the log and drop the puma thread commit

carbonin:
Yeah, fairly sure metrics starts its own webserver entirely separate from puma. [ref](https://github.com/RedHatInsights/insights-api-common-rails/blob/6c308db7d77d16201442a7687bc761fd05d7dbe1/lib/insights/api/common/metrics.rb#L19) and also [the implementation uses a webrick server](https://github.com/discourse/prometheus_exporter/blob/b8fc6838b09cb9c91421466ea8285fc9b7cb8118/lib/prometheus_exporter/server/web_server.rb#L39-L43)

bdunne:
Sorry, I forgot about that, the other Prometheus exporter that I was looking at uses the puma server

PR #150, opened 2020-01-09
Move openapi.json files to just x.y version numbers

agrare:
Make it easier to tag the .z version of the API without having to create
new openapi.json files everytime.


agrare:
Hm I thought https://github.com/RedHatInsights/sources-api/pull/149 should have fixed this hakiri warning

PR #151, opened 2020-01-10
Added a timeout to the application availability checks performed.

abellotti:
- Setting both open and read timeout to 10 seconds.


abellotti:
there you go @agrare, not sure of the value though.

agrare:
10 seconds seems like a reasonable place to start

PR #152, opened 2020-01-13
Update source types static icon assets url.

Hyperkid123:
### Changes
- added/updated URLs to chrome static assets

### DIFF
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -55,10 +55,11 @@ amazon:
         :initializeOnMount: true
         :initialValue: aws
   :vendor: Amazon
-  :icon_url: "/aws_logo.png"
+  :icon_url: "/apps/chrome/assets/images/partners-icons/aws.svg"
 ansible-tower:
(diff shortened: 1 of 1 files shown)


db/seeds/source_types.yml
```diff
@@ -113,6 +114,7 @@ ansible-tower:
 azure:
   :product_name: Microsoft Azure
   :vendor: Azure
+  :icon_url: "/apps/chrome/assets/images/partners-icons/miscrosoft-azure.svg"
```
slemrmartin: typo?

Hyperkid123: Yep that gotta be fixed in chrome though. I will take care of that, but it will take a while.

slemrmartin: :+1: ok

Hyperkid123: Pr is here: https://github.com/RedHatInsights/insights-chrome/pull/547
It will take time though until a new version of chrome is released

PR #157, opened 2020-01-23
[WIP] StatusController for Liveness Probe

slemrmartin:
TPINVTRY-827
Health check endpoint for OpenShift Liveness Probe

---

* [ ] **depends on**: https://github.com/RedHatInsights/insights-api-common-rails/pull/143 (and depends on publishing it to RubyGems)


carbonin:
This has to wait for a released version of the gem, right? Probably should mark it as WIP until then.

slemrmartin:
I marked it with **depends on** checkbox, but I'll add wip too

PR #158, opened 2020-01-23
Add logging to the seeding step

agrare:
To help improve debugging of missing seeded records we can log
creates/updates/deletes for the seed classes.


agrare:
QE was seeing missing azure source_types which should be seeded, this should help debug what is going on

PR #159, opened 2020-01-24
Don't share a kafka connection

agrare:
It seems that sharing a common kafka connection for posting to different
topics causes issues with being unable to assign partitions to messages
leading to Kafka::DeliverFailed errors

### DIFF
--- a/lib/sources/api/events.rb
+++ b/lib/sources/api/events.rb
@@ -12,7 +12,18 @@ module Sources
 
         publish_opts[:headers] = headers if headers
 
-        Messaging.client.publish_topic(publish_opts)
+        messaging_client.publish_topic(publish_opts)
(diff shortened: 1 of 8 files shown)


carbonin:
I think it would be good to have a better name for the messaging client that isn't the event one.

Looking at the code it feels like events is special rather than the reality which is we need a separate connection per topic. Maybe the thread local key should be named after the intended topic?

Anyway, I can merge this now since we need to test this as a fix in CI. Sound goo do you @agrare ?

agrare:
> Looking at the code it feels like events is special rather than the reality which is we need a separate connection per topic. Maybe the thread local key should be named after the intended topic?

Yeah I was actually considering this long term if we confirm this fixes things

PR #160, opened 2020-01-24
Go back to a single puma thread

agrare:
A single puma thread seems to be the only case where everything was working properly


agrare:
cc @slemrmartin @carbonin 

@carbonin I'm just going to go back to a different connection for both of these to get things back to a known working state

PR #162, opened 2020-01-24
Check list of available topics before publishing a message

agrare:
To prevent messages destined for non-existant topics blocking other valid messages from being delivered, check the list of topics (cached) before publishing a message.

### DIFF
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -39,7 +39,7 @@ module Api
 
           logger.debug("publishing message to #{topic}...Complete")
         else
-          logger.debug("not publishing message to non-existing topic: #{topic}")
+          logger.error("not publishing message to non-existing topic: #{topic}")
         end
 
         check_application_availability(source)
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -23,19 +23,24 @@ module Api
         source = Source.find(params[:source_id])
         topic  = "platform.topological-inventory.operations-#{source.source_type.name}"
 
-        logger.debug("publishing message to #{topic}...")
-        messaging_client = Sources::Api::Messaging.client(topic)
-        messaging_client.publish_topic(
-          :service => topic,
-          :event   => "Source.availability_check",
-          :payload => {
-            :params => {
-              :source_id       => source.id.to_s,
-              :external_tenant => source.tenant.external_tenant
+        if Sources::Api::Messaging.topics.include?(topic)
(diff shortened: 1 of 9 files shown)


lib/sources/api/messaging.rb
```diff
@@ -1,16 +1,24 @@
...
         require "manageiq-messaging"
 
-        @client ||= {}
-        @client[topic] ||= ManageIQ::Messaging::Client.open(
-          :protocol => :Kafka,
-          :host     => ENV["QUEUE_HOST"] || "localhost",
-          :port     => ENV["QUEUE_PORT"] || "9092",
-          :encoding => "json"
-        )
+        Thread.current[:messaging_client] ||= begin
```
agrare: Went back to caching per thread but want to leave this at 1 puma thread first to test one variable at a time

carbonin:
So this is now a single client per thread (of which we only have one) as opposed to one per topic, right?
But in this case we're ensuring that we don't try to publish to topics that don't exist?

app/controllers/api/v1/sources_controller.rb
```diff
@@ -23,19 +23,24 @@ def check_availability
...
+              }
             }
-          }
-        )
-        logger.debug("publishing message to #{topic}...Complete")
+          )
+
+          logger.debug("publishing message to #{topic}...Complete")
+        else
+          logger.debug("not publishing message to non-existing topic: #{topic}")
```
carbonin: Ideally this shouldn't happen ... can we make this an error or warn?

agrare:
> So this is now a single client per thread (of which we only have one) as opposed to one per topic, right?
> But in this case we're ensuring that we don't try to publish to topics that don't exist?

@carbonin yes exactly

app/controllers/api/v1/sources_controller.rb
agrare: So the way the availability checker works right now this actually happens rather often, and not just for source-types that we don't have an operations worker for yet.

E.g. There are the cloudforms source-types like ovirt, vsphere, openstack where I don't think we'll ever have an operations worker.

agrare: I think the way to properly fix this, is to turn this current check into the topological-inventory application check.  So sources that aren't enabled for topology don't try to write to this topic.  Then topology is playing at the same level as all the other apps on the platform.

carbonin: Yeah, I was thinking this was happening pretty often, but we log debug by default right now anyway so I was thinking making this an error would make the failing topics easier to find.

agrare: :+1: done

PR #165, opened 2020-01-29
Allow read-only access for graphql

slemrmartin:
although grapqhl request is POST, it's actually like GET so RBAC read-only access should be allowed

### DIFF
--- a/app/controllers/api/v1/graphql_controller.rb
+++ b/app/controllers/api/v1/graphql_controller.rb
@@ -21,6 +21,13 @@ module Api
         )
         render :json => result
       end
+
+      private
+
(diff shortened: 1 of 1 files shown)


slemrmartin:
cc @abellotti

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -21,6 +21,13 @@ def query
         )
         render :json => result
       end
+
+      private
+
+      # RBAC readonly access is allowed for graphql's POST
+      def request_is_readonly
+        true
```
agrare: We'll need to remember to update this when we allow graphql to be used for create/update

Issue #172, opened 2020-02-04
Add explicit relationship between application and authentication


infinitewarp:
As I understand the current model design, for a given source, an application _could_ have multiple authentications, and an authentication _could_ belong to multiple applications. As a client of Sources API (and its Kafka topics), I should be able to look at a given auth and know precisely what apps are relevant for it, and vice versa, but that does not currently appear to be possible.

Here's an example breakdown of this problem:

- I have a source with a single endpoint and two applications.
- Each application has a _different_ authentication under that same endpoint.
- Both authentications have the same auth _type_.
- How does a Sources API client know which auth to use for which app?

One real-world use case here is with an AWS Source and the Cost Management and Subscription Watch (coming soon) apps. Both of those apps may have an ARN-type auth, but they will have different auth instances because the associated IAM roles will have significantly different policies in AWS.

My dev team spoke with @agrare about this yesterday (2020-02-03), and it seemed that adding some kind of foreign key relationship in Sources would be a reasonable solution to this problem. Until we have a solution to the aforementioned problem, Cloud Meter's integration for Subscription Watch may be blocked.

PR #173, opened 2020-02-05
Add a Cloud Meter application and Auth Type

agrare:
This adds an application_type for CloudMeter and also adds a new
cloud-meter-arn auth type.  Since the ARN that cloud meter needs is
different from the one that cost management needs we need a different
auth-type for now.

Fixes https://github.com/RedHatInsights/sources-api/issues/172

### DIFF
--- a/db/seeds/application_types.yml
+++ b/db/seeds/application_types.yml
@@ -22,6 +22,14 @@
     - tenant_id_client_id_client_secret
     openshift:
     - token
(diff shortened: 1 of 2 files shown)


agrare:
Hey @rvsia can you check this format for the DDF schema? Thanks!

agrare:
Thanks @rvsia !

PR #177, opened 2020-02-10
Add Tagging to sources

agrare:
Add tags and tagging intermediate tables to sources

Depends on: https://github.com/RedHatInsights/insights-api-common-rails/pull/158

Relates to JIRA: TPINVTRY-815


app/models/authentication_tag.rb
```diff
@@ -0,0 +1,4 @@
+class AuthenticationTag < ApplicationTag
```
bdunne: This should be `< ApplicationRecord`

app/models/endpoint_tag.rb
```diff
@@ -0,0 +1,4 @@
+class EndpointTag < ApplicationTag
```
bdunne: Same here

app/models/source_tag.rb
```diff
@@ -0,0 +1,4 @@
+class SourceTag < ApplicationTag
```
bdunne: And here

app/models/concerns/act_as_taggable_on.rb
```diff
@@ -0,0 +1,60 @@
+module ActAsTaggableOn
```
bdunne: What do you think of moving this into InsightsApiCommon?

bdunne:
Any tests?

agrare:
Nope not yet

Issue #183, opened 2020-02-17
As a user I want Sources to support a shared ARN authentication between different applications


werwty:
This is probably a future enhancement.

As things are now, when an Authentication is created it's type `arn` for cost management, and `cloud-meter-arn` for Cloud Meter. It is very possible that a user would have a singular ARN that has the policy configured to support but Cost Management and Cloud Meter. 

I would expect an ARN authentication to be application agnostic. But for there to be some foreign key relationship typing an ARN to specific Applications. That way an ARN can belong to Cost Management, Cloud Meter, both, or some other future application with ARN support.

agrare:
Okay backend for this is merged, now we just need the UI to start setting this when configuring the source.

Going to assign this to @rvsia for now

rvsia:
Not sure, what I am supposed to do. Just to add another API call (create authapp) when creating a source with an application?

However, there are more UI concerns I have. If it's possible to have multiple authentication of the same type, then we need to add some authentication management to Sources UI.

Users should be able to
- remove authentication from application and use different one
- add a new authentication, not assigned to an app (?)
- choose an existing authentication or create a new authentication when attaching an app to an existing source
- see which applications are using an authentication when editing the authentication (in the edit form)

agrare:
@rvsia I believe that is what they are asking for (unless @werwty intends to do this all from the API).  We should start with at a minimum creating this link when we create an authentication record the way we do today so that it can be looked up more efficiently and then we'll need UX input on designing an authentication manager

Issue #182, opened 2020-02-17
Creating 2+ applications and 2+ authentications can lead to ambiguity.


werwty:
The UI disallows the creation of multiple authentications of the same type. But this is doable via the API.

Steps to reproduce:
```
## Create a Source
http -a insights-qa POST https://ci.cloud.redhat.com/api/sources/v1.0/sources name=cloudmeter-demo source_type_id=2

## Create an Endpoint
http -a insights-qa POST https://ci.cloud.redhat.com/api/sources/v1.0/endpoints role=aws default=true source_id=1575


## Create two Application
http -a insights-qa POST https://ci.cloud.redhat.com/api/sources/v1.0/applications application_type_id=5 source_id=1575
http -a insights-qa POST https://ci.cloud.redhat.com/api/sources/v1.0/applications application_type_id=5 source_id=1575

## Create two Authentications
http -a insights-qa POST https://ci.cloud.redhat.com/api/sources/v1.0/authentications authtype=cloud-meter-arn password=arn:aws:iam::114204391493:role/cloudigrade-ci-role resource_type=Endpoint resource_id=1093

http -a insights-qa POST https://ci.cloud.redhat.com/api/sources/v1.0/authentications authtype=cloud-meter-arn password=arn:aws:iam::114204391493:role/cloudigrade-ci-role-2 resource_type=Endpoint resource_id=1093

```

There is no way of figuring out which Authentication belongs to which Application. This ambiguity is problematic when we try to figure out whether an authentication should be deleted when a 'cloud-meter` application is deleted.

agrare:
Hey @werwty the addition of the cloud meter auth type was a stopgap until we got a hard link between the app and the auth.  I started that work here https://github.com/RedHatInsights/sources-api/pull/184 but for now using the authtype is the best we have.

werwty:
:+1: Good to know! I didn't know that was in progress. Thanks @agrare !

slemrmartin:
done

PR #184, opened 2020-02-17
Add an application-authentication join table

agrare:
This is needed to resolve a couple of issues:
https://github.com/RedHatInsights/sources-api/issues/183
https://github.com/RedHatInsights/sources-api/issues/182

which revolve around not having a direct link between an application and an authentication.

### DIFF
--- a/app/models/application_authentication.rb
+++ b/app/models/application_authentication.rb
@@ -2,4 +2,6 @@ class ApplicationAuthentication < ApplicationRecord
   include TenancyConcern
   belongs_to :application
   belongs_to :authentication
+
+  before_validation(:on => :create) { |i| i.send("tenant=", i.authentication.send("tenant")) }
 end
--- a/app/models/application_authentication.rb
+++ b/app/models/application_authentication.rb
@@ -1,0 +1,5 @@
+class ApplicationAuthentication < ApplicationRecord
+  include TenancyConcern
+  belongs_to :application
+  belongs_to :authentication
+end
--- a/db/migrate/20200217181820_add_application_authentications.rb
+++ b/db/migrate/20200217181820_add_application_authentications.rb
@@ -1,0 +1,11 @@
+class AddApplicationAuthentications < ActiveRecord::Migration[5.2]
+  def change
+    create_table :application_authentications do |t|
+      t.references :tenant,         :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :application,    :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :authentication, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.timestamps
+      t.index %w[tenant_id application_id authentication_id], :unique => true, :name => "index_on_tenant_application_authentication"
+    end
+  end
+end
--- a/spec/requests/api/v2.0/application_authentications_spec.rb
+++ b/spec/requests/api/v2.0/application_authentications_spec.rb
@@ -107,5 +107,17 @@ RSpec.describe("v2.0 - ApplicationAuthentications") do
         )
       end
     end
+
+
+    context "delete" do
+      it "success: with a valid paylod" do
+        record = ApplicationAuthentication.create!(payload)
+
+        delete(instance_path(record.id), :headers => headers)
+
+        expect(response.status).to eq(204)
+        expect(response.parsed_body).to be_empty
+      end
+    end
   end
 end
--- a/spec/requests/api/v2.0/application_authentications_spec.rb
+++ b/spec/requests/api/v2.0/application_authentications_spec.rb
@@ -1,6 +1,12 @@
 RSpec.describe("v2.0 - ApplicationAuthentications") do
   include ::Spec::Support::TenantIdentity
 
+  let(:messaging_client) { instance_double("ManageIQ::Messaging::Client") }
+  before do
+    allow(messaging_client).to receive(:publish_topic)
+    allow(Sources::Api::Messaging).to receive(:client).and_return(messaging_client)
+  end
+
   let(:headers)          { {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
   let(:source_type)      { SourceType.create!(:name => "my-source-type", :product_name => "My Source Type", :vendor => "ACME") }
   let(:source)           { Source.create!(:source_type => source_type, :tenant => tenant, :uid => SecureRandom.uuid, :name => "my-source") }
@@ -10,6 +16,12 @@ RSpec.describe("v2.0 - ApplicationAuthentications") do
   let(:application)      { Application.create!(:application_type => application_type, :source => source, :tenant => tenant) }
   let(:attributes)       { {"application_id" => application.id.to_s, "authentication_id" => authentication.id.to_s } }
   let(:collection_path)  { "/api/v2.0/application_authentications" }
+  let(:payload) do
+    {
+      "application_id"    => application.id.to_s,
+      "authentication_id" => authentication.id.to_s
+    }
+  end
 
   describe("/api/v2.0/application_authentication") do
     context "get" do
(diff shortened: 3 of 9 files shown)


agrare:
WIP until I can add tests for the `application/:id/authentications` subcollection

agrare:
@bdunne the idea here is to add a hard link between an application and a specific authentication record so that apps don't have to guess based on application.auth_type they can do `GET /applications/:id/authentications`

I bumped v1.0 to v1.0.1 but didn't bump v2.0 because there are still a couple of things I think we want to do before releasing v2.0

agrare:
Wondering if we should do a data migration to set this for existing sources, thoughts?

bdunne:
> I bumped v1.0 to v1.0.1

I think this change is minimally a minor version bump since we're adding a new feature (a primary and subcollection).  Should we just roll it up into v2.0?

> there are still a couple of things I think we want to do before releasing v2.0

Is there a list?  If this feature is needed soon, should we release v2.0 with these changes and push the rest of the to-do list to v3?

> Wondering if we should do a data migration to set this for existing sources, thoughts?

Probably a good idea to back-fill the existing relationships.

agrare:
> I think this change is minimally a minor version bump since we're adding a new feature (a primary and subcollection). Should we just roll it up into v2.0?

That's fair, I'll just add it to v2 only I don't think it's worth carrying v1, v1.1, and v2

agrare:
> Is there a list? If this feature is needed soon, should we release v2.0 with these changes and push the rest of the to-do list to v3?

The tenant change was one, another I wanted to look at was changing the polymorphic belongs_to on authentication to a straight belongs_to endpoint since it doesn't need to belong to e.g. hosts in sources and it makes creating them in the API harder but that can wait for another major since the tenant change is already in.

slemrmartin:
@agrare I would like to understand what impact this PR will have. Is it replacement of Source-Authentication relation or extension? How it'll behave in UI/orchestrator? I think whole team meeting will be great

agrare:
Hey @slemrmartin the linked issues describe the problem pretty well.

App teams want to be able to find an authentication record that they are intended to use without looping through and trying to match up based on what authtype they're looking for.

It isn't a replacement for anything so the orchestrator shouldn't be affected.

agrare:
@bdunne can you review the new migration?

db/migrate/20200217181820_add_application_authentications.rb
```diff
@@ -0,0 +1,11 @@
+class AddApplicationAuthentications < ActiveRecord::Migration[5.2]
+  def change
+    create_table :application_authentications do |t|
+      t.references :tenant,         :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :application,    :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
```
bdunne: Will this delete the `application` if the `application_authentication` is deleted?

spec/requests/api/v2.0/application_authentications_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  let(:source_type)      { SourceType.create!(:name => "my-source-type", :product_name => "My Source Type", :vendor => "ACME") }
+  let(:source)           { Source.create!(:source_type => source_type, :tenant => tenant, :uid => SecureRandom.uuid, :name => "my-source") }
+  let(:application_type) { ApplicationType.create!(:name => "my-app") }
+  let(:endpoint)         { Endpoint.create!(:source => source, :tenant => tenant) }
+  let(:authentication)   { Authentication.create!(:tenant => tenant, :resource => endpoint) }
+  let(:application)      { Application.create!(:application_type => application_type, :source => source, :tenant => tenant) }
+  let(:attributes)       { {"application_id" => application.id.to_s, "authentication_id" => authentication.id.to_s } }
+  let(:collection_path)  { "/api/v2.0/application_authentications" }
+
+  describe("/api/v2.0/application_authentication") do
```
bdunne: `create`, `update` and `delete` tests would be nice too if you have some time.

app/models/application_authentication.rb
```diff
@@ -0,0 +1,5 @@
+class ApplicationAuthentication < ApplicationRecord
```
bdunne: I think we need a `before_validation` hook to set the tenant also.

bdunne: A request spec covering `POST /api/v2.0/application_authentication` should cover that

db/migrate/20200217181820_add_application_authentications.rb
agrare: No it will delete this record if the application is deleted

app/models/application_authentication.rb
agrare: :+1:

agrare: Added https://github.com/RedHatInsights/sources-api/pull/184/files#diff-aca3e3260b70e2940cde506fd8bf2f7fR6 and added https://github.com/RedHatInsights/sources-api/pull/184/files#diff-2b415817a335813d2379e808df1b8593R49

werwty:
@agrare Any updates on this? It'll be great if this change could make it in soon, so we can start using it for cloudigrade.

agrare:
That's up to @bdunne

db/migrate/20200217235235_populate_application_authentications.rb
```diff
@@ -0,0 +1,41 @@
...
+      next if auth.resource_type != "Endpoint"
+
+      endpoint = Endpoint.find(auth.resource_id)
+      next if endpoint.nil?
+
+      source_type_name = endpoint.source&.source_type&.name
+
+      applications = endpoint.source&.applications || []
+      applications.each do |application|
+        next if application.tenant&.id.nil?
```
bdunne: Do we need a `Tenant` stub here?

db/migrate/20200217235235_populate_application_authentications.rb
```diff
@@ -0,0 +1,41 @@
...
+  def up
+    Authentication.all.each do |auth|
+      next if auth.resource_type != "Endpoint"
+
+      endpoint = Endpoint.find(auth.resource_id)
+      next if endpoint.nil?
+
+      source_type_name = endpoint.source&.source_type&.name
+
+      applications = endpoint.source&.applications || []
```
bdunne: Do we need an `Application` stub?

db/migrate/20200217235235_populate_application_authentications.rb
```diff
@@ -0,0 +1,41 @@
...
+      applications.each do |application|
+        next if application.tenant&.id.nil?
+
+        application_type = application.application_type
+        next if application_type.nil?
+
+        supported_auth_types = application_type.supported_authentication_types[source_type_name] || []
+        if supported_auth_types.include?(auth.authtype)
+          application.application_authentications.create!(:authentication_id => auth.id, :tenant_id => application.tenant.id)
```
bdunne: And `ApplicationAuthentications`

infinitewarp:
Any updates? Just bumping because we have other work blocked in cloudigrade pending resolution of #183 and #182.

agrare:
@bdunne ptal

db/migrate/20200217235235_populate_application_authentications.rb
bdunne: `Application` needs a `belongs_to :tenant` or we can use `application.tenant_id` to avoid instantiating a Tenant object

db/migrate/20200217235235_populate_application_authentications.rb
```diff
@@ -0,0 +1,47 @@
...
+      endpoint = Endpoint.find(auth.resource_id)
+      next if endpoint.nil?
+
+      source_type_name = endpoint.source&.source_type&.name
+
+      applications = endpoint.source&.applications || []
+      applications.each do |application|
+        next if application.tenant&.id.nil?
+
+        application_type = application.application_type
```
bdunne: This relation isn't defined in the stub model

db/migrate/20200217235235_populate_application_authentications.rb
```diff
@@ -0,0 +1,47 @@
...
+        next if application_type.nil?
+
+        supported_auth_types = application_type.supported_authentication_types[source_type_name] || []
+        if supported_auth_types.include?(auth.authtype)
+          application.application_authentications.create!(:authentication_id => auth.id, :tenant_id => application.tenant.id)
```
bdunne: Same here

PR #185, opened 2020-02-17
Don't return the 'tenant' attribute

agrare:
The tenant attribute maps to the user's identity.account_number which is only relevant internally.  Users of the API who login through 3scale will only ever be able to see a single tenant so this attribute isn't helpful and can be confusing.  Also since it isn't a proper relationship (it is the tenant.external_tenant not tenant_id) it can trip up things like filtering.  In v2.0 we should drop this attribute.

Relates to JIRA: TPINVTRY-742

### DIFF
--- a/lib/tasks/openapi_generate.rake
+++ b/lib/tasks/openapi_generate.rake
@@ -1,32 +1,5 @@
 require 'insights/api/common/open_api/generator'
 class OpenapiGenerator < Insights::API::Common::OpenApi::Generator
-  def generator_blacklist_allowed_attributes
-    @generator_blacklist_allowed_attributes ||= {
-      :tenant_id => ['Authentication', 'Application', 'Endpoint', 'Source'].to_set.freeze
-    }
-  end
-
-  def generator_blacklist_substitute_attributes
-    @generator_blacklist_substitute_attributes ||= {
-      :tenant_id => ["tenant", { "type" => "string", "readOnly" => true }].freeze
-    }
-  end
-
-  def schemas
-    @schemas ||= begin
-      super.merge(
-        "Tenant" => {
-          "type"       => "object",
-          "properties" => {
-            "name"            => {"type" => "string", "readOnly" => true, "example" => "Sample Tenant"},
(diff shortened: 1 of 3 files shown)


agrare:
@bdunne removing this from the openapi doc seemed to be the best way to keep it in the v1 responses but not in the v2 ones.

Once we remove the v1 API we can remove the as_json override in the tenancy mixin.

lib/tasks/openapi_generate.rake
```diff
@@ -1,32 +1,5 @@
 require 'insights/api/common/open_api/generator'
 class OpenapiGenerator < Insights::API::Common::OpenApi::Generator
-  def generator_blacklist_allowed_attributes
-    @generator_blacklist_allowed_attributes ||= {
-      :tenant_id => ['Authentication', 'Application', 'Endpoint', 'Source'].to_set.freeze
-    }
-  end
```
agrare: I started looking into having this stick around for v1 but not v2 and found that `Insights::API::Common::OpenApi::Generator` only updates the latest API anyway.  This means that to update an older API you have to comment out all newer APIs in `config/routes.rb`.  Not great but I'm not sure that I want to solve that right now.

spec/requests/api/v2.0/sources_spec.rb
```diff
@@ -134,16 +134,6 @@
           :parsed_body => Insights::API::Common::ErrorDocument.new.add(400, "Record not unique").to_h
         )
       end
-
-      it "ignores blacklisted params" do
```
gtanzillo: Is this still needed to test https://github.com/RedHatInsights/sources-api/blob/master/app/controllers/application_controller.rb#L92?

Or, does it fail on the post because "tenant" was removed from the schema?

agrare: Right this fails sooner because tenant isn't even defined as an attribute.

This test still exists in the v1.0 spec

lib/tasks/openapi_generate.rake
bdunne: Yeah, maybe we should allow for an ENV variable to select the version you want to run the generator for.

PR #189, opened 2020-02-24
Handle invalid routes better

lpichler:
as done here before https://github.com/RedHatInsights/topological_inventory-api/pull/188 

cc @slemrmartin

### DIFF
--- a/app/controllers/api.rb
+++ b/app/controllers/api.rb
@@ -1,2 +1,11 @@
 module Api
+  class RootController < ApplicationController
+    skip_before_action :validate_primary_collection_id
+    skip_before_action :validate_request
+
+    def invalid_url_error
+      error_document = Insights::API::Common::ErrorDocument.new.add(404, "Invalid URL /#{params[:path]} specified.")
+      render :json => error_document.to_h, :status => 404
+    end
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -100,4 +100,6 @@ Rails.application.routes.draw do
       resources :tenants,         :only => [:index, :show]
     end
   end
+
+  match "*path", :to => "api/root#invalid_url_error", :via => ActionDispatch::Routing::HTTP_METHODS - [:head]
 end
--- a/spec/requests/general_spec.rb
+++ b/spec/requests/general_spec.rb
@@ -1,0 +1,26 @@
+RSpec.shared_examples "invalid_url_requests" do |request_type, invalid_url, http_method|
+  include ::Spec::Support::TenantIdentity
+
+  let(:headers) { {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
+
+  context "with invalid #{request_type} URLs" do
+    it "fail with an error object" do
+      send(http_method, invalid_url, :headers => headers)
+
+      if http_method == :head
+        expect(response.status).to eq(404)
+      else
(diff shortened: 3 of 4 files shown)


app/controllers/api.rb
```diff
@@ -1,2 +1,10 @@
 module Api
+  class RootController < ApplicationController
+    skip_before_action :validate_primary_collection_id
+
```
slemrmartin: @lpichler POST requests raised some error, please try

lpichler: fixed, thanks 👍

lpichler:
@abellotti we want to [return the proper error message for invalid URLs](https://github.com/RedHatInsights/topological_inventory-api/pull/188) as you did before.

and there is issue with HEAD request:

1.  when you do HEAD request and there is route in `routes.rb`
 ```ruby
match "*path", :to => "api/root#invalid_url_error", :via =>ActionDispatch::Routing::HTTP_METHODS
```
=> **request will end in root controller**


2.  when you do HEAD request and there is **not present** route from point 1.  in `routes.rb`
=> **request will end in related resource controller** (this is expected)

So in second commit, I added code to handle HEAD request in root controller and check if we have such route in GET and it will fails in not. 

What do you think about it ?

thanks!

**Comment by abellotti** [comment]
Finally got around to testing this a bit, sorry about the delay @lpichler. for an active record not found, while the 404 is properly returned, I'm getting the "Invalid URL /api/v2.0/sources/123 specified." which is  misleading. it's a correct URL, error message should just be "Not found" or "Record not found". Thanks

File: app/controllers/api.rb [hunk 2]
```diff
@@ -1,2 +1,18 @@
...
+    skip_before_action :validate_primary_collection_id
+    skip_before_action :validate_request
+
+    def head_request_exists_in_get_routes?
+      routes = Rails.application.routes.recognize_path(request.path, :method => :get)
+      routes.extract!(:controller, :action) != {:controller => "api/root", :action => "invalid_url_error"}
+    end
+
+    def invalid_url_error
+      return head :ok if request.head? && head_request_exists_in_get_routes?
```
bdunne: Should this ever render `:ok`?  We only get here is the url is invalid, right?

syncrou: @lpichler would you be able to provide any followup to @bdunne?

lpichler: I removed logic for `HEAD` request because it need to solved differently - maybe by adding head routes to `routes.rb`

lpichler: I removed logic for HEAD and we need to investigate whether there is better way how to handle HEAD request.

lpichler:
> Finally got around to testing this a bit, sorry about the delay @lpichler. for an active record not found, while the 404 is properly returned, I'm getting the "Invalid URL /api/v2.0/sources/123 specified." which is misleading. it's a correct URL, error message should just be "Not found" or "Record not found". Thanks

@abellotti 
we are using in specs format

`/api/v1.0/sources/123`

but it looks like that correct url is:

`/api/sources/v1.0/sources` and `/api/sources/v1.0/` is prefix.

then `/api/v2.0/sources/123` is not correct and this `/api/sources/v2.0/sources/123` returns not found as expected.

@slemrmartin which format is correct ?

lpichler:
Ready to re-review 👍 

thanks

slemrmartin:
@lpichler is this PR still active?

PR #193, opened 2020-02-28
Fix bug with graphql posts with non-org-admin

agrare:
Fix a bug introduced by renaming request_is_readonly to request_is_readonly? (since it is a boolean method) [here](https://github.com/RedHatInsights/sources-api/pull/190/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eL68) but this broke the graphql controller which overrides this to allow for posts to be treated as readonly.

Also add a spec test to cover this case.

### DIFF
--- a/spec/requests/api/graphql_controller_spec.rb
+++ b/spec/requests/api/graphql_controller_spec.rb
@@ -20,5 +20,14 @@ RSpec.describe("v1.0 - GraphQL") do
       expect(response.status).to eq(200)
       expect(result_source_tenant(response.body)).to match_array([tenant.external_tenant])
     end
+
+    it "with non-org-admin: returns the external tenant identifier" do
+      headers = { "CONTENT_TYPE" => "application/json", "x-rh-identity" => non_org_admin_identity }
+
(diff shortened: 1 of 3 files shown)


spec/requests/api/graphql_controller_spec.rb
```diff
@@ -10,7 +10,7 @@ def result_source_tenant(response_body)
   end
 
   context "querying source tenant" do
-    before { stub_const("ENV", "BYPASS_TENANCY" => nil) }
+    #before { stub_const("ENV", "BYPASS_TENANCY" => nil) }
```
agrare: Not sure why this is here but it passes without it?

agrare: I'll handle this in a separate PR

slemrmartin: I think old code tested presence of key in ENV

agrare: Oh it looks like it is turned on/off as a `before {}` block but never restored so this might have been a sporadic test failure "fix"

PR #204, opened 2020-03-25
Enabling filtering by association attribute for the REST API.

abellotti:
- [x] Depends on https://github.com/RedHatInsights/insights-api-common-rails/pull/159


slemrmartin:
@abellotti can you post some examples? Better to readme of insights-api-common. Thanks

abellotti:
README.md just merged there, so https://github.com/RedHatInsights/insights-api-common-rails/blob/master/README.md  now has the examples.

bdunne:
@abellotti Can you also require the latest insights-api-common v3.9?

PR #207, opened 2020-03-26
Trackable Source#availability_check

slemrmartin:
Unifying log messages

Availability can be traced in Kibana by message `Source#availability_check <id>`


slemrmartin:
@gtanzillo @carbonin together with https://github.com/RedHatInsights/sources-monitor/pull/10 it can be now traced with one common message. Starting and Finishing message have the same format in all operations

slemrmartin:
cc @lpichler

PR #211, opened 2020-04-15
Update DDF strings to version 2

rvsia:
There is a new major DDF version which changes some strings. Please see https://data-driven-forms.org/migration-guide

- removes unnecessary `validator` naming from validator types
- `switch-field` to `switch`

Should be merged after 
- [x] https://github.com/RedHatInsights/sources-ui/pull/315

The UI has fallbacks prepared.

### DIFF
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -32,8 +32,8 @@ amazon:
         :label: S3 bucket name
         :isRequired: true
         :validate:
-        - :type: required-validator
-        - :type: pattern-validator
+        - :type: required
+        - :type: pattern
           :pattern: "^[A-Za-z0-9]+[A-Za-z0-9_-]*$"
       - :name: authentication.password
         :stepKey: arn
@@ -41,10 +41,10 @@ amazon:
         :label: ARN
         :isRequired: true
         :validate:
-        - :type: required-validator
-        - :type: pattern-validator
+        - :type: required
+        - :type: pattern
           :pattern: "^arn:aws:.*"
-        - :type: min-length-validator
+        - :type: min-length
           :threshold: 10
     - :type: cloud-meter-arn
       :name: Subscription Watch ARN
@@ -59,10 +59,10 @@ amazon:
         :label: ARN
         :isRequired: true
         :validate:
-        - :type: required-validator
-        - :type: pattern-validator
+        - :type: required
+        - :type: pattern
           :pattern: "^arn:aws:.*"
-        - :type: min-length-validator
+        - :type: min-length
           :threshold: 10
     :endpoint:
       :hidden: true
@@ -93,14 +93,14 @@ ansible-tower:
         :label: User name
         :isRequired: true
         :validate:
-        - :type: required-validator
+        - :type: required
       - :component: text-field
         :name: authentication.password
         :label: Secret key
         :type: password
         :isRequired: true
         :validate:
-        - :type: required-validator
+        - :type: required
     :endpoint:
       :title: Configure Ansible Tower endpoint
       :fields:
@@ -113,8 +113,8 @@ ansible-tower:
         :name: url
         :label: URL
         :validate:
-        - :type: url-validator
-      - :component: switch-field
+        - :type: url
+      - :component: switch
         :name: endpoint.verify_ssl
         :label: Verify SSL
       - :component: text-field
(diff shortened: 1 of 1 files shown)


rvsia:
The CI failure does not seem related to this PR

syncrou:
@lindgrenj6 are we going to need a migration for this for existing data?

lindgrenj6:
@syncrou yes, because we need to update the seeds that are already there. I can do that quick if you would like @rvsia

rvsia:
Has something been changed? Because we have never had to migrate the seeds before. 🤔 

But yeah, please, do it. :)

db/migrate/db_migrate_20200513201952_rename_source_type_ddf_strings_v2_rb.rb
```diff
@@ -0,0 +1,60 @@
+class RenameSourceTypeDdfStringsV2 < ActiveRecord::Migration[5.2]
```
lindgrenj6, line 1: I'd rename this file to `20200513201952_rename_source_type_ddf_strings_v2.rb` that way it follows the rail standard

db/migrate/db_migrate_20200513201952_rename_source_type_ddf_strings_v2_rb.rb
```diff
@@ -0,0 +1,60 @@
...
+  def down
+    rename(DDF_V2_MAPPINGS.map(&:reverse))
+  end
+
+  private
+
+  def rename(mappings)
+    SourceType.all.each do |st|
+      schema = st.schema.try(:with_indifferent_access)
+      next unless schema
```
syncrou, line 23: Is there any way that if `schema` isn't created and we skip it, that we skip a record we needed? 

Should we bail completely if `schema.nil?`

lindgrenj6: I think it's just normal - there are some SourceTypes that don't have schemas, this is on CI:

```
irb(main):001:0> SourceType.where(:schema => nil).map(&:product_name)
  SourceType Load (0.5ms)  SELECT "source_types".* FROM "source_types" WHERE "source_types"."schema" IS NULL
{"@timestamp":"2020-05-14T20:33:40.148967 ","hostname":"sources-api-946-njsm6","pid":66580,"tid":"2afa9f5545b4","level":"debug","message":"  SourceType Load (0.5ms)  SELECT \"source_types\".* FROM \"source_types\" WHERE \"source_types\".\"schema\" IS NULL"}
=> ["Red Hat OpenStack", "Red Hat Virtualization", "Red Hat CloudForms", "VMware vSphere"]
```

Lines up with the `seeds.rb` file:
OpenStack:
https://github.com/RedHatInsights/sources-api/blob/a4c2e9368e91e3b7458dd002fc7c419193a9ecac/db/seeds/source_types.yml#L220-L223
RHEV:
https://github.com/RedHatInsights/sources-api/blob/a4c2e9368e91e3b7458dd002fc7c419193a9ecac/db/seeds/source_types.yml#L224-L227
CF:
https://github.com/RedHatInsights/sources-api/blob/a4c2e9368e91e3b7458dd002fc7c419193a9ecac/db/seeds/source_types.yml#L170-L172
vSphere:
https://github.com/RedHatInsights/sources-api/blob/a4c2e9368e91e3b7458dd002fc7c419193a9ecac/db/seeds/source_types.yml#L259-L262

syncrou: Cool. Thanks.

db/migrate/db_migrate_20200513201952_rename_source_type_ddf_strings_v2_rb.rb
rvsia: uppps, sorry! :facepalm:

gmcculloug:
@syncrou @lindgrenj6 I believe these changes will be applied by the seeding logic: https://github.com/RedHatInsights/sources-api/blob/master/app/models/concerns/seeding_concern.rb#L17

And the migration is not needed at all.

syncrou:
@gmcculloug @rvsia - I didn't get that the seeding logic ran at startup. I now see how it would solve the problem of bad key names.

@lindgrenj6 lets pull out the migration then.

PR #216, opened 2020-05-01
Pin common gem to 3.9 to prevent spec failures

lindgrenj6:


Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
-gem 'insights-api-common',  '~> 3.9'
+gem 'insights-api-common',  '=  3.8'
```
syncrou, line 6: Hoping the `~> 3.9` wasn't for any features only available from 3.9 on.

syncrou: Looking at the failing specs we need at least v3.9 to deal with the GraphQL changes.

lindgrenj6: Pinning to 3.9 looks like its working!

PR #218, opened 2020-05-05
Add 'Red Hat' to OpenShift product name, drop CloudForms

rvsia:
OpenShift logo contains a little bit different naming than the product name:

OpenshiftLogo

vs

`OpenShift Container Platform`

@syncrou This should be probably decided by the copy, but I am not sure who is it. :D 

Mocks

https://marvelapp.com/574481j

### DIFF
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -171,7 +171,7 @@ cloudforms:
   :product_name: Red Hat CloudForms
   :vendor: Red Hat
 openshift:
-  :product_name: OpenShift Container Platform
+  :product_name: Red Hat OpenShift Container Platform
   :schema:
     :authentication:
     - :type: token
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -167,9 +167,6 @@ azure:
         :hideField: true
         :initializeOnMount: true
         :initialValue: azure
-cloudforms:
-  :product_name: Red Hat CloudForms
-  :vendor: Red Hat
 openshift:
   :product_name: Red Hat OpenShift Container Platform
   :schema:


rvsia:
From this discussion https://github.com/RedHatInsights/sources-api/pull/217#issuecomment-624118052 it seems that the Ansible Tower was right and the icon was wrong.

rvsia:
I removed the Ansible Tower from this PR. (The Icon is wrong, we have to update the icon - not the product_name)

Feel free to close it, if this is not right.

gmcculloug:
@sbuenafe-rh Can you help on this or mention someone that can?  Thanks.

sbuenafe-rh:
> @sbuenafe-rh Can you help on this or mention someone that can? Thanks.

@KendraMar @rpg1216

gmcculloug:
@KendraMar @rpg1216 Do you have any thoughts/comments on this PR?

I agree with the change and thing we should merge it.  But I also want to suggest that if we are aligning Product names we should address "Ansible Tower" which would be changed to "Red Hat Ansible Tower".

I realize this does not totally match the image currently being display for that Source type, but that battle has already been fought. 

```diff
ansible-tower:
-  :product_name: Ansible Tower
+  :product_name: Red Hat Ansible Tower
  :vendor: Red Hat
```

cc @syncrou

rvsia:
@gmcculloug Should I add Red Hat Ansible Tower here too?

Also, I should probably write a migration. :thinking:

gmcculloug:
@rvsia No need for a migration, there is seeding logic that will apply the changes: https://github.com/RedHatInsights/sources-api/blob/master/db/seeds.rb

Let's modify `Ansible Tower` to `Red Hat Ansible Tower` and also remove the `cloudforms` section as that is no longer needed.  Also, please update the PR title and description to take these changes into account.

rvsia:
@gmcculloug Okay. Will do it!

I have a question, in the #211 we made a migration for the `schema` attribute. What's the difference here? :thinking:

gmcculloug:
@rvsia I think the seeding will take care of the changes for PR #211 as well and the migration is not needed.  I'll comment on that PR.

Don't forgot to drop the cloudforms section from the file:
```
cloudforms:
  :product_name: Red Hat CloudForms
  :vendor: Red Hat
```

rvsia:
@gmcculloug Yeah, we have never did the migration before, so I was confused by that requirement.

KendraMar:
@gmcculloug Yup, good catch. This branding stuff is agony.

(no logo) Red Hat Ansible Tower

rvsia:
@KendraMar @gmcculloug from the last update in the email, it seems that Ansible Tower is just Ansible Tower without any Red Hat. 🤷‍♂️

KendraMar:
@rvsia @gmccullough

Yeah, just read that from Bill. My two cents: as long as we're identifying that Ansible Tower is the thingy you connect to, I think we need to rely on PM and TMM to police branding/logo use. I'm not sure how much more UX and dev can do.

gmcculloug:
@rvsia I'm replying to the email.  Let's try and get agreement there and then update this PR.

gmcculloug:
@rvsia Drop the "Red Hat" on Ansible Tower and :shipit:

PR #219, opened 2020-05-07
Support for status type of string

mkanoor:
This is needed to conform with the IPP which requires that error objects conform to the JSON API.

Uses the newer version of the common gem.

https://jsonapi.org/format/#errors
Which states that status is a string and not an integer.

https://projects.engineering.redhat.com/browse/TPINVTRY-925

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -53,13 +53,13 @@ class ApplicationController < ActionController::API
           ActsAsTenant.without_tenant { yield }
         end
       rescue KeyError, Insights::API::Common::IdentityError
-        error_document = Insights::API::Common::ErrorDocument.new.add("401", 'Unauthorized')
+        error_document = Insights::API::Common::ErrorDocument.new.add('401', 'Unauthorized')
         render :json => error_document.to_h, :status => error_document.status
       rescue Insights::API::Common::EntitlementError
-        error_document = Insights::API::Common::ErrorDocument.new.add("403", 'Forbidden')
+        error_document = Insights::API::Common::ErrorDocument.new.add('403', 'Forbidden')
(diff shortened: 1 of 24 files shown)


mkanoor:
@syncrou Please review

app/controllers/application_controller.rb
```diff
@@ -12,29 +12,29 @@ class RbacError < StandardError; end
   around_action :with_current_request
 
   rescue_from ActiveRecord::RecordNotFound do |exception|
```
lindgrenj6, line 14: I'm ok with these `rescue_from`'s for now since they were already here, but we should probably move these to an exception mapping class and let the common gem handle it like we do for catalog/approval.

I'll make an issue for it.

app/controllers/application_controller.rb
```diff
@@ -53,13 +53,13 @@ def with_current_request
           ActsAsTenant.without_tenant { yield }
         end
       rescue KeyError, Insights::API::Common::IdentityError
-        error_document = Insights::API::Common::ErrorDocument.new.add(401, 'Unauthorized')
+        error_document = Insights::API::Common::ErrorDocument.new.add("401", 'Unauthorized')
         render :json => error_document.to_h, :status => error_document.status
       rescue Insights::API::Common::EntitlementError
-        error_document = Insights::API::Common::ErrorDocument.new.add(403, 'Forbidden')
+        error_document = Insights::API::Common::ErrorDocument.new.add("403", 'Forbidden')
```
syncrou, line 59: The smallest of nits.  Can we be consistent with quotes? Not mixing '' with ""

mkanoor:
@syncrou Please review, I have fixed the quotes

PR #222, opened 2020-05-14
[WIP] Remove ansible tower icon

rvsia:
It seems that Ansible Tower should not have any icon/logo at all.

cc @gmcculloug @syncrou 

@lindgrenj6 Can you please check the migration? I have never did it before :smile_cat:

### DIFF
--- a/db/migrate/20200514132558_remove_ansible_tower_icon.rb
+++ b/db/migrate/20200514132558_remove_ansible_tower_icon.rb
@@ -1,0 +1,7 @@
+class RemoveAnsibleTowerIcon < ActiveRecord::Migration[5.2]
+  def change
+    ansible_tower = SourceType.find_by(name: 'ansible-tower')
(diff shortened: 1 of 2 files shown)


gmcculloug:
@rvsia I am trying to follow what you are suggesting with this PR.

Is this a reaction to the email thread that stated:

> We only have one product logo. 
> 
> Red Hat Ansible Automation Platform. 
> 
> Tower logo should not be used. 

This says to me that we should continue to use the logo we currently have in place.

rvsia:
@gmcculloug 

Yes, it is a reaction to the thread, however It says to me that we should not use any logo for Ansible Tower. :D

gmcculloug:
I'm going to close this PR for now.  If anything we might end up with a Tower based image that does not show any icon, meaning a text-only image, to replace the current one.  For now we need to continue with the image we currently have in place.

PR #227, opened 2020-05-21
Add nullable true to availability_status

syncrou:
JIRA: https://projects.engineering.redhat.com/browse/TPINVTRY-940

Testing with the new open api generator is resulting in expected behavior on POST for attributes without `nullable: true`.

### DIFF
--- a/public/doc/openapi-3-v3.0.json
+++ b/public/doc/openapi-3-v3.0.json
@@ -1442,7 +1442,8 @@
             "$ref": "#/components/schemas/ID"
           },
           "availability_status": {
-            "type": "string"
+            "type": "string",
+            "nullable": true
           },
           "availability_status_error": {
             "type": "string"


syncrou:
@slemrmartin that's a good question. This came up in discussions with Yuval.  There is a QE test that was passing in `availability_status` as nil. Version 3 generated open api client was correctly failing on that POST because the `availability_status`attribute does not allow nils.  

Yuval was surprised he ran into the error in the first place, as previous client versions had allowed them.

Catalog ran into these issues last summer. Catalog / Approval now have delta checking between the openapi validation in api-comon and the openapi json spec included in the repo. My guess is there will be a number of these issues popping up in sources moving forward.

We can potentially confinue this conversation in JIRA: https://projects.engineering.redhat.com/browse/TPINVTRY-940

slemrmartin:
@syncrou I wrote the arguments to the Jira. 

Anyway whether we decide to allow `availability_status` to be nullable or not,
`availability_status_error` fields (Application, Endpoint, Authentication) should be nullable in all cases.

syncrou:
@slemrmartin in this [comment](https://github.com/RedHatInsights/sources-api/pull/227#issuecomment-637479272) are you suggesting I added nullable true to the other attributes as well in the sources-api openapi spec?

slemrmartin:
@syncrou yes

syncrou:
@slemrmartin - Looking into this more: There is no need to set nullable true on availability_status.

I believe the QE tests should not pass in a null availability_status from their tests. I'm going to close this.

slemrmartin:
@syncrou ah, I thought you'll add the 'nullable' for error messages there. 
Implemented in https://github.com/RedHatInsights/sources-api/pull/231

PR #229, opened 2020-05-29
Raise Kafka events when source and its associated applications are deleted

hsong-rh:
https://projects.engineering.redhat.com/browse/TPINVTRY-916

The logic of raising Kafka messages originally is defined in `application controller`'s destroy method (via `mixin`), so it will be missed even when `:dependent => :destroy` is set. This PR moved it into `after_destroy` callback, to guarantee message is sent out.

### DIFF
--- a/app/models/concerns/event_concern.rb
+++ b/app/models/concerns/event_concern.rb
@@ -1,0 +1,14 @@
+module EventConcern
+  extend ActiveSupport::Concern
+
+  included do
+    after_destroy :raise_event, :prepend => true
+  end
+
+  def raise_event
+    headers = Insights::API::Common::Request.current_forwardable
+    logger.debug("publishing message to topic \"platform.sources.event-stream\"...")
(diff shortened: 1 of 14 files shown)


hsong-rh:
@slemrmartin Thank you, this is exactly what I want to discuss about. Since the Jira ticket only mentioned `applications`, I am not sure whether other resources need the same fix.

slemrmartin:
@hsong-rh there should be involved all current receivers of these events

app/models/concerns/event_concern.rb
```diff
@@ -0,0 +1,14 @@
+module EventConcern
+  extend ActiveSupport::Concern
+
+  included do
+    after_destroy :raise_event, :prepend => true
```
lindgrenj6, line 5: This is cool, I had no idea about `:prepend => true` to make sure the dependents are deleted before the parent.

PR #231, opened 2020-06-05
Adding nullable:true to availability_status_error

slemrmartin:
`availability_status_error` message needs to be nullable, because the message should be deleted when Source becomes available. 

---

[TPINVTRY-948](https://projects.engineering.redhat.com/browse/TPINVTRY-948)

### DIFF
--- a/public/doc/openapi-3-v3.0.json
+++ b/public/doc/openapi-3-v3.0.json
@@ -1445,6 +1445,7 @@
             "type": "string"
           },
           "availability_status_error": {
+            "nullable": true,
             "type": "string"
           },
           "created_at": {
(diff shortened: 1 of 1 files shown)


slemrmartin:
**related**: https://github.com/RedHatInsights/insights-api-common-rails/pull/198

public/doc/openapi-3-v3.0.json
```diff
@@ -1445,6 +1445,7 @@
             "type": "string"
           },
           "availability_status_error": {
+            "nullable": true,
```
syncrou, line 1448: I don't think we need this @slemrmartin. Setting it to empty string should be fine.

Unless allowing it to be `nil` is better than `''` in your opinion?

slemrmartin: I didn't realize this is an option :) You're right

slemrmartin:
closing

PR #232, opened 2020-06-10
Application check

roliveri:
Validate source compatibility when creating Application.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -17,6 +17,6 @@ class Application < ApplicationRecord
 
   def source_must_be_compatible
   	return if application_type.supported_source_types.include?(source.source_type.name)
-  	errors.add(:source, "is not compatible with this application type")
+  	errors.add(:source, "of type: #{source.source_type.name}, is not compatible with this application type")
   end
 end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -10,4 +10,13 @@ class Application < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
+
+  validate :source_must_be_compatible
+
+  private
+
+  def source_must_be_compatible
+  	return if application_type.supported_source_types.include?(source.source_type.name)
+  	errors.add(:source, "is not compatible with this application type")
+  end
 end
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe("Application") do
       it "should raise RecordInvalid" do
         expect do
           subject
-        end.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Source is not compatible with this application type")
+        end.to raise_error(ActiveRecord::RecordInvalid, /^.* is not compatible with this application type/)
       end
     end
   end
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -16,15 +16,17 @@ RSpec.describe("Application") do
     }
   end
 
-  describe "Application" do
-    context "create!" do
-      it "Should return an instance of Application on success" do
+  describe "create!" do
+    context "when the application supports the given source type" do
+      it "should return an instance of Application" do
         expect(
           Application.create!(attributes.merge(:source_id => compatible_source.id))
         ).to be_an_instance_of(Application)
       end
+    end
 
-      it "Should raise RecordInvalid when given an incompatible source" do
+    context "when the application does not support the given source type" do
+      it "should raise RecordInvalid" do
         expect do
           Application.create!(attributes.merge(:source_id => incompatible_source.id))
         end.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Source is not compatible with this application type")
--- a/spec/models/application_spec.rb
+++ b/spec/models/application_spec.rb
@@ -1,0 +1,34 @@
+RSpec.describe("Application") do
+  include ::Spec::Support::TenantIdentity
+
+  let(:compatible_source)         { Source.create!(:source_type => compatible_source_type, :tenant => tenant, :uid => SecureRandom.uuid, :name => "my-source") }
+  let(:compatible_source_type)    { SourceType.create!(:name => "my-source-type", :product_name => "My Source Type", :vendor => "ACME") }
+
+  let(:incompatible_source)       { Source.create!(:source_type => incompatible_source_type, :tenant => tenant, :uid => SecureRandom.uuid, :name => "not-my-source") }
+  let(:incompatible_source_type)  { SourceType.create!(:name => "not-my-source-type", :product_name => "Not My Source Type", :vendor => "ACME") }
+
+  let(:application_type)  { ApplicationType.create!("name" => "my-application", :supported_source_types => ["my-source-type"]) }
+
+  let(:attributes) do
+    {
+      :application_type_id => application_type.id,
+      :tenant              => tenant
+    }
+  end
+
+  describe "Application" do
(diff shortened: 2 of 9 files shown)


spec/requests/api/v1.0/sources_spec.rb
```diff
@@ -451,10 +451,12 @@ def check_availability_path(source_id)
         source      = Source.create!(attributes.merge("tenant" => tenant))
 
         app_type1 = ApplicationType.create(:name         => "/platform/application-type1",
-                                           :display_name => "Application Type One")
+                                           :display_name => "Application Type One",
+                                           :supported_source_types => ["openshift"])
```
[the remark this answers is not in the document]
slemrmartin: Don't know about any historical reason

spec/models/application_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:application_type)  { ApplicationType.create!("name" => "my-application", :supported_source_types => ["my-source-type"]) }
+
+  let(:attributes) do
+    {
+      :application_type_id => application_type.id,
+      :tenant              => tenant
+    }
+  end
+
+  describe "Application" do
```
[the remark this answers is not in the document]
roliveri: The extra `describe "Application"` was unintentional, thanks for catching that.
As for the rest, I agree.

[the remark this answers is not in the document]
roliveri: I would volunteer to tackle transitioning to FactoryBot in a follow-up PR. It would have saved a lot of work in this case.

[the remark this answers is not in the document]
roliveri: @lindgrenj6 @eclarizio - Updated.

spec/models/application_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+
+    context "when the application does not support the given source type" do
+      it "should raise RecordInvalid" do
+        expect do
+          Application.create!(attributes.merge(:source_id => incompatible_source.id))
+        end.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Source is not compatible with this application type")
+      end
+    end
```
eclarizio, line 34: It's a little bit weird that we have these contexts, but there's no setup for them that indicates why they matter as contexts.

After looking through the rest of the spec, I understand what's going on, but I think this whole file can actually be refactored a bit to use a `let` for either the source_type name or for the supported source types.

Given the code is checking for inclusion, it seems to make sense that a `let` for the `supported_source_types` should be used instead, and keep the `source_type_name` as an inline "my-source-type". Maybe something like this?:
```ruby
RSpec.describe("Application") do
  include ::Spec::Support::TenantIdentity

  describe "create!" do
    let(:source) { Source.create!(:source_type => source_type, :tenant => tenant, :uid => SecureRandom.uuid, :name => "my-source") }
    let(:source_type) { SourceType.create!(:name => "my-source-type", :product_name => "My Source Type", :vendor => "ACME") }
    let(:application_type)  { ApplicationType.create!("name" => "my-application", :supported_source_types => supported_source_types) }

    subject do
      Application.create!({:application_type_id => application_type.id, :tenant => tenant, :source_id => source.id})
    end

    context "when the application supports the given source type" do
      let(:supported_source_types) { ["my-source-type", "my-other-source-type-not-present-here"] }

      it "returns an instance of Application" do
        expect(subject).to be_an_instance_of(Application)
      end
    end

    context "when the application does not support the given source type" do
      let(:supported_source_types) { ["my-other-source-type-not-present-here"] }

      it "raises a RecordInvalid exception" do
        expect do
          subject
        end.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Source is not compatible with this application type")
      end
    end
  end
end
```

gmcculloug:
@roliveri If you rebase you'll pick up #233 which resolves the Hakiri failure so all tests are green.

spec/models/application_spec.rb
roliveri: Won't the `let(:application_type)` under the `describe "create!"` be memoized on the first reference, so changing `supported_source_types` won't have any effect on the second reference?

I also think additional test cases are needed for `Application`, that aren't specific to this PR. It would probably make more sense to refactor this spec file after that is fleshed out.

eclarizio: @roliveri No, it's lazily loaded for each run-through, and since you would have another lazily loaded `let` for the `supported_source_types`, that gets evaluated at runtime for those specific contexts. I think even if the `let(:application_type)` was `let!(:application_type)` (so, not lazily loaded), due to the fact that you have the `let(:supported_source_types)` within the different contexts means depending on the `it` block that RSpec is currently working on, it should pull in the correct `let`.

roliveri: @eclarizio - Actually, I think it would be clearer if we change the `source` in the set-up for the context, rather than the `supported_source_types`.

We would have the same `subject` set up, and just:
```
let(:source) {  compatible_source }
```
for one context, and:
```
let(:source) {  incompatible_source }
```
for the other.

I prefer this because, it's easier to follow overall, and it more directly illustrates what's being tested - the same application with different sources. As opposed to the same source, and changing the application indirectly.

eclarizio: @roliveri That's probably fine, but from the standpoint of the spec, every run through of an `it` block, it *is* a different source already, even though it's referencing the same `let`. I would suggest though to change the context descriptions to reflect that setup, though. It seems odd to say "this is what happens when the application doesn't support X" but the setup doesn't change the application at all.

Like I said, this is kind of subjective, and I just have a lot of opinions built up over the years of working the way I suggested and why I like it the best. I don't want to give the impression you guys *have* to follow what I say 😄

roliveri: @ eclarizio - I think the context descriptions are fine the way they are. True, the application isn't changing between the contexts, but the source is - it shows the application supporting one source but not the other. While it's true, technically it's a different instantiation of the source, but they're instantiated in the same exact way - so they're functionally identical, that's what's significant. The difference between the sources is made clear through the variable names.

I guess it's all how you look at it.

I don't mind subjective suggestions. I don't have much experience writing spec tests, so it's good to get input from those that have. I'll make an update based on my last comments

eclarizio: > While it's true, technically it's a different instantiation of the source, but they're instantiated in the same exact way - so they're functionally identical, that's what's significant.

Right, which is why to me, it makes sense to have one source let, and you only adjust the one thing that matters, the source_type_name. Again, nothing wrong with the way you're doing it, it still ends up working, it's mainly in the name of isolating "what sole thing important for this context?". This helps when your specs rely on many different factors, because if you have to keep nesting contexts because you have a ton of different variables, it allows you to see the cyclomatic complexity from the specs where it may have been kind of "hidden" in the code. Then you know you should probably look at pulling some logic out or refactoring into a separate class or something similar.

roliveri: Like I said, it's all in how you look at it. I don't happen agree with this aspect of your argument as it's applied to this specific situation.

app/models/application.rb
```diff
@@ -10,4 +10,13 @@ class Application < ApplicationRecord
...
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
+
+  validate :source_must_be_compatible
+
+  private
+
+  def source_must_be_compatible
+  	return if application_type.supported_source_types.include?(source.source_type.name)
+  	errors.add(:source, "is not compatible with this application type")
```
syncrou, line 20: Maybe we should include the name of the source here? "`#{source.source_type.name}` is not compatible with this application type"

roliveri: That would be useful. I'll add it.

PR #233, opened 2020-06-11
Update puma versioning resolve Hakiri warnings

gmcculloug:
Following version from https://github.com/RedHatInsights/topological_inventory-api/blob/master/Gemfile#L16


slemrmartin:
@gmcculloug @syncrou these hakiri "security" improvements are potentially dangerous, better to open a bug on some Hakiri page if possible

gmcculloug:
@slemrmartin The Hakiri support is through email which I have used in the past and they has been very responsive.  I'll reach out to them, thanks.

PR #234, opened 2020-06-18
Added byebug to Gemfile

mkanoor:
This allows us to step thru code in production by having a simple
script and runnning it with bin/rails r <script_name>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
+gem 'byebug'
 gem 'insights-api-common',  '~> 4.0'
 gem 'jbuilder',             '~> 2.0'
 gem 'json-schema',          '~> 2.8'


mkanoor:
@syncrou @lindgrenj6 
Please review.

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
+gem 'byebug'
```
lindgrenj6, line 6: Can we move this to the `:development` section? That is where we have it for catalog - and it's still there in production if we need it.

```
[ ⚛ ] crc $ k exec -it catalog-api-259-wbghj -- gem list | grep byebug
byebug (11.1.3)
```

mkanoor: @lindgrenj6 That doesn't make sense I think we are doing "bundle install" in a dev environment and moving that code into production without doing a "bundle update" in QA or Prod. The way the code is promoted from Dev => QA => Production might be saving us here. If we ran CI in production mode we would not see that gem.

syncrou: @mkanoor I have to agree with @lindgrenj6 on this.  I would rather that byebug wasn't in our deployed `Gemfile` in the main context. 

Also, if we really did add a byebug and attach to the console from prod - any production request hitting that pod would be stuck as well.  

I'm thinking there is a better OCP solution to this without including a debugger on a production environment.

mkanoor: @syncrou This is not adding byebug in the main code, you drop a new ruby script which has the byebug in it and that's the one that allows you to step thru the code without changing anything. We had encountered this in catalog and the the thing that helped was having a ruby script with the byebug. A sample script would look like

```
require 'byebug'
order_number = 283
o = Order.find(order_number).order_items.first
req = {:headers => o.context['headers'],
       :original_url => o.context['original_url']}
byebug
Insights::API::Common::Request.current = req
Thread.current[:user_context] = UserContext.new(Insights::API::Common::Request.current, {})
access = Thread.current[:user_context].access
scopes = access.scopes('portfolios','order', 'catalog')
ActsAsTenant.with_tenant(Tenant.find_by(:external_tenant => Insights::API::Common::Request.current.tenant)) do
  o = Order.find(order_number)
  op = OrderPolicy.new(Thread.current[:user_context], o)
  response = op.submit_order?
  puts response
end
```

So your main line production code has no changes.

We have already implemented this in catalog and approval and is the only way to debug right in production env with production data.
https://github.com/RedHatInsights/approval-api/pull/327

gmcculloug: If we do add this we should use `:require => false`

> Specify :require => false to prevent bundler from requiring the gem, but still install it and maintain dependencies.

https://bundler.io/gemfile.html

lindgrenj6: @mkanoor I guess my main argument here is that we don't really _need_ byebug to run the application. The nice thing about OCP is we could set up a different image that is built off of stable but with the "debug" tools in the Gemfile.

From there we could spin up just a pod (like this: https://github.com/lindgrenj6/insights-rds-db-dump-container/blob/master/catalog/resources/catalog_restorer.yml) with the `source-api:debug` image, it would connect to the db exactly like the regular sources-api pod but it would have all the tools necessary baked in. This way we wouldn't have to include the debugging tools in the main image. 

-----

I'm working on a script that you just give two arguments:
1) the container image (in this case it would be `sources-api:debug` 
2) the deploymentconfig to pull the ENV settings from, that way it mounts the necessary secrets

From there is runs a pod and drops you in a shell so one can debug safely in a disposable environment.

Once this is done it would make it really easy to have debugging pods that we spin up when necessary rather than including them in the prod image build.

mkanoor:
Messed up in my repo locally I will close this one and open a new one.

PR #236, opened 2020-06-18
Log the stack trace for KeyError and IdentityError

mkanoor:


app/controllers/application_controller.rb
```diff
@@ -55,7 +55,8 @@ def with_current_request
         else
           ActsAsTenant.without_tenant { yield }
         end
-      rescue KeyError, Insights::API::Common::IdentityError
+      rescue KeyError, Insights::API::Common::IdentityError => e
+        logger.debug(e.backtrace.join("\n"))
```
gmcculloug, line 59: Two suggestions:
1. Move this backtrace after the line below so the stack trace follows the error message.
2. Use `logger.debug { e.backtrace.join("\n") }` where the block will only be evaluated when we are actually going to log the message.

Also, I'm curious, but not asking to change in this PR, why don't we log the actual error message instead of the `Key Error or Insights::API::Common::IdentityError` message.  Seems we could actually specify which issue was encountered.

mkanoor: @gmcculloug Its all being done here
https://github.com/RedHatInsights/insights-api-common-rails/blob/bc964c7ec61028e311215b5df29d368c2ca30333/lib/insights/api/common/application_controller_mixins/exception_handling.rb#L17
I am trying to debug something related to why playbooks are getting a 401.
We would need another rework of the controller to remove all the rescue froms and delegate   it to the common gem and it does all the work we need, including the exception name and the stack trace. This is just temporary

PR #239, opened 2020-06-19
Add availability_status_listener subscribing to Kafka

bzwei:
https://projects.engineering.redhat.com/browse/TPINVTRY-970

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -1,0 +1,78 @@
+require "manageiq-messaging"
+
+class AvailabilityStatusListener
+  SERVICE_NAME = "platform.sources.status".freeze
+  GROUP_REF = "sources-api-status-worker".freeze
+  EVENT_AVAILABILITY_STATUS = "availability_status".freeze
+
+  attr_accessor :messaging_client_options, :client
+
+  def initialize(messaging_client_options = {})
+    self.messaging_client_options = default_messaging_options.merge(messaging_client_options)
+  end
+
+  def run
+    Thread.new { subscribe_to_availability_status }
+  end
+
+  def subscribe_to_availability_status
+    Rails.logger.info("Sources API availability_status listener started...")
+
+    ManageIQ::Messaging::Client.open(messaging_client_options) do |client|
+      client.subscribe_topic(
+        :service     => SERVICE_NAME,
+        :persist_ref => GROUP_REF,
+        :max_bytes   => 500_000
+      ) do |event|
+        process_event(event)
+      end
+    end
+  rescue => e
+    Rails.logger.error(["Something is wrong with Kafka client: ", e.message, *e.backtrace].join($RS))
+    retry
+  end
+
+  private
+
+  def process_event(event)
+    Rails.logger.info("Kafka message #{event.message} received with payload: #{event.payload}")
+    return unless event.message == EVENT_AVAILABILITY_STATUS
+
+    model_class = event.payload["resource_type"].classify.constantize
+    validate_resource_type(model_class)
+    record_id = event.payload["resource_id"]
+    object = model_class.find(record_id)
+    options = {
+      :availability_status       => event.payload["status"],
+      :availability_status_error => event.payload["error"],
+      :last_checked_at           => Time.now.utc
+    }
+    options[:last_available_at] = options[:last_checked_at] if options[:availability_status] == 'available'
+
+    object.update!(options)
+    raise_event("#{model_class}.update", object.as_json)
+  rescue NameError
+    Rails.logger.error("Invalid resource_type #{event.payload["resource_type"]}")
+  rescue ActiveRecord::RecordNotFound
+    Rails.logger.error("Could not find #{model_class} with id #{record_id}")
+  rescue ActiveRecord::RecordInvalid
+    Rails.logger.error("Invalid status #{event.payload["status"]}")
+  rescue => e
+    Rails.logger.error(["Something is wrong when processing Kafka message: ", e.message, *e.backtrace].join($RS))
+  end
+
+  def raise_event(event, payload)
+    Sources::Api::Events.raise_event(event, payload)
+  rescue => e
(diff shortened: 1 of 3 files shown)


bzwei:
A few notes:
1. codeclimate reports 2 issues, both complaining rescue `Exception`. The main purpose of rescuing all possible exceptions is to prevent the listener thread from crashing. Should we allow the thread to die while the pod itself keeps on running? I also have some concerns about rescuing and restarting the messaging client on error. It may keep on restarting if the underlying error cannot be resolved soon.
2. Both `Application` and `Endpoint` models have `last_checked_at` and `last_available_at` attributes but seems not updated by the API controller. Should the listener automatically update them?
3. API controller update logic always publishes a Kafka message to `platform.sources.event-stream` on every update. Should this new listener do the same thing?

bzwei:
@slemrmartin @syncrou please review

syncrou:
@bzwei I don't have an issue with Hakiri complaint. To keep the Thread stable I'd rather rescue `Exception`.

syncrou:
>Both Application and Endpoint models have last_checked_at and last_available_at attributes but seems not updated by the >API controller. Should the listener automatically update them?

My understanding is that those are updated by the `availability_status` check in the subsequent collector?  @lindgrenj6 is that correct?  Also does the collector only update the `Endpoint` availability_status but not the `Application` availability_status? That is my 'current' understanding of how it works.

syncrou:
>API controller update logic always publishes a Kafka message to platform.sources.event-stream on every update. Should >this new listener do the same thing?

If we could I would like to keep the new listener on the same topic.

lindgrenj6:
>My understanding is that those are updated by the availability_status check in the subsequent collector? @lindgrenj6 is that correct? 

@syncrou yep, that is correct. It's all right here: 
https://github.com/lindgrenj6/topological_inventory-providers-common/blob/5c04f92462815485bffcb4f798bb82a490e28a82/lib/topological_inventory/providers/common/operations/source.rb#L107-L110

https://github.com/lindgrenj6/topological_inventory-providers-common/blob/5c04f92462815485bffcb4f798bb82a490e28a82/lib/topological_inventory/providers/common/operations/source.rb#L90-L92

>Also does the collector only update the Endpoint availability_status but not the Application availability_status? 

That is what it looks like, it updates the `Source` and `Endpoint`. I don't see any reference to `Application` in the availability check operation in the providers-common gem.

lib/availability_status_listener.rb
```diff
@@ -0,0 +1,66 @@
...
+    Rails.logger.error("Invalid resource_type #{event.payload["resource_type"]}")
+  rescue ActiveRecord::RecordNotFound
+    Rails.logger.error("Could not find #{model_class} with id #{record_id}")
+  rescue ActiveRecord::RecordInvalid
+    Rails.logger.error("Invalid status #{event.payload["status"]}")
+  rescue => e
+    Rails.logger.error(["Something is wrong when processing Kafka message: ", e.message, *e.backtrace].join($RS))
+  end
+
+  def validate_resource_type(model_class)
```
lindgrenj6, line 58: I like this method - but we aren't using it anywhere it appears. Maybe we could be checking on line 41? 

Also we should have `Source` in there as well since that needs updated.

bzwei:
> > API controller update logic always publishes a Kafka message to platform.sources.event-stream on every update. Should >this new listener do the same thing?
> 
> If we could I would like to keep the new listener on the same topic.

I have requested to create new topic `platform.sources.status`. My original question was should the listener send a new Kafka message to `platform.sources.event-stream` once it receives an update event from the status topic? Also it seems odd if we listen and publish to the same topic.

bzwei:
@lindgrenj6 @syncrou From the source code you referenced it seems the sender should provide `last_checked_at` and `last_available_at`. It makes more sense to me that we make them read-only and automatically updated by `sources-api`.

lindgrenj6:
@bzwei I agree, maybe we could include them in the `object.update` call on L44. 

It would only update the `last_available_at` when `event.payload["status"]` was `"available"`, but `last_checked_at` would always be the current timestamp. 

Is that something like what you're thinking?

syncrou:
@lindgrenj6 @bzwei - After our offline conversations.  In answer to question 3 from Bill: Yes

Also, for this change we do not need versioning since the original behavior of the API is intact.  Moving forward when `last_checked_at` and `last_available_at` are made read only, we will make a version change.

bzwei:
@syncrou @lindgrenj6 All addressed. I also tested that topic `platform.sources.status` already created at CI env.

lib/availability_status_listener.rb
```diff
@@ -0,0 +1,78 @@
...
+        :service     => SERVICE_NAME,
+        :persist_ref => GROUP_REF,
+        :max_bytes   => 500_000
+      ) do |event|
+        process_event(event)
+      end
+    end
+  rescue => e
+    Rails.logger.error(["Something is wrong with Kafka client: ", e.message, *e.backtrace].join($RS))
+    retry
```
syncrou, line 32: Does `retry` come from the `manageiq-messaging` gem?

bzwei: Retry will start another messaging client and subscribe it to the topic again.

gmcculloug: `retry` is a core Ruby concept: https://ruby-doc.org/docs/keywords/1.9/Object.html#method-i-retry

slemrmartin:
@syncrou is it intended that we switch all topological availability checkers to use kafka instead of API calls?

lindgrenj6:
@slemrmartin As far as I understand it, yes. The platform wants to move away from using internal api calls. We also want to mark the `availability_status` fields as `readOnly` so the users cannot update them. 

@syncrou please clarify if I missed anything.

gmcculloug:
@slemrmartin Yes, that is the intention.  Using the internal API requires the applications to modify (hack) the insights header to elevate privileges.  By switching to use internal Kafka messaging the callers will not be required to make this change.

PR #241, opened 2020-07-06
Update messaging gem version 0.1.6

bzwei:


Gemfile
```diff
@@ -7,7 +7,7 @@ gem 'insights-api-common',  '~> 4.0'
 gem 'jbuilder',             '~> 2.0'
 gem 'json-schema',          '~> 2.8'
 gem 'manageiq-loggers',     '~> 0.4', ">= 0.4.2"
-gem 'manageiq-messaging',   '~> 0.1.5', :require => false
+gem 'manageiq-messaging',   '~> 0.1.6', :require => false
```
gmcculloug, line 10: LGTM, just wanted to ask if the `:require => false` is needed?

Seems we could remove about 12 calls to `require "manageiq-messaging"` if we remove this flag.

syncrou: @bzwei do you have any thoughts on ☝️

syncrou: My 2 cents.  We tend to fall back on `require` false for dependency loading performance reasons. It is sometimes less effencient codewise, but more efficient run time wise.

PR #240, opened 2020-06-23
Making authtype readonly(OpenAPI)

pkomanek:
OpenAPI definitions: Add `readOnly` for authtype in Authentication

Based on https://projects.engineering.redhat.com/browse/TPINVTRY-932.

Issue #196, opened 2020-03-03
PATCHing the authtype should not be allowed


werwty:
Currently if I create an Authentication with authtype `arn`, I would expect that the authtype to be non update-able. 

But the following PATCH request is allowed, and does update my authtype to azure: 
```http -a insights-qa PATCH https://ci.cloud.redhat.com/api/sources/v1.0/authentications/1139 authtype=azure```

agrare:
@slemrmartin can you make authtype a readOnly attribute in the openapi definition?

slemrmartin:
sorry, I didn't notice this issue, will do

slemrmartin:
TPINVTRY-932

gmcculloug:
@slemrmartin @agrare The PR for this change was merged already, but I wanted to point out that we likely only wanted to make it read-only for the PATCH call and other actions like create.  The UI is getting away with this because it makes direct API calls, but anyone using the OpenAPI generated client would fail on a create now that this field is read-only.

cc @zsadeh @lindgrenj6 @syncrou

slemrmartin:
@pkomanek Can you make a follow up PR with the change @gmcculloug described above? (PR is https://github.com/RedHatInsights/sources-api/pull/240)

PR #242, opened 2020-07-08
Make Authentication:authtype readonly only for updating

slemrmartin:
**Issue**: #196 

We want to mark the authtype as readOnly only for PATCH request (update). As the openapi generator isn't prepared for it yet, this is a fast solution for it

---

Reverting: #240 

---

[TPINVTRY-932](https://projects.engineering.redhat.com/browse/TPINVTRY-932)

### DIFF
--- a/app/models/authentication.rb
+++ b/app/models/authentication.rb
@@ -11,4 +11,12 @@ class Authentication < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
+
+  validate :authtype_not_updated
+
+  def authtype_not_updated
+    if !new_record? && authtype_changed?
+      errors.add(:authtype, "cannot be updated")
+    end
+  end
 end
--- a/lib/tasks/openapi_generate.rake
+++ b/lib/tasks/openapi_generate.rake
@@ -31,11 +31,12 @@ class OpenapiGenerator < Insights::API::Common::OpenApi::Generator
     end
   end
 
-  def schema_overrides
-    authentication = schemas['Authentication']
-    authentication['properties']['authtype']['readOnly'] = true
-    super.merge('Authentication' => authentication)
-  end
+  # TODO: make readOnly authtype only for update
+  # def schema_overrides
+  #   authentication = schemas['Authentication']
+  #   authentication['properties']['authtype']['readOnly'] = true
+  #   super.merge('Authentication' => authentication)
+  # end
 end
 
 namespace :openapi do
--- a/app/controllers/api/v1/mixins/update_mixin.rb
+++ b/app/controllers/api/v1/mixins/update_mixin.rb
@@ -4,6 +4,8 @@ module Api
       module UpdateMixin
         def update
           record = model.update(params.require(:id), params_for_update)
+          raise ActiveRecord::RecordInvalid.new(record) if record.invalid?
+
           raise_event("#{model}.update", record.as_json)
           head :no_content
         end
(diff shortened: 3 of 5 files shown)


slemrmartin:
cc @gmcculloug @pkomanek 

@lindgrenj6 if you have an idea how to integrate it to the openapi.json, please share it there for the following PR

lib/tasks/openapi_generate.rake
```diff
@@ -31,11 +31,12 @@ class OpenapiGenerator < Insights::API::Common::OpenApi::Generator
...
-    authentication = schemas['Authentication']
-    authentication['properties']['authtype']['readOnly'] = true
-    super.merge('Authentication' => authentication)
-  end
+  # TODO: make readOnly authtype only for update
+  # def schema_overrides
+  #   authentication = schemas['Authentication']
+  #   authentication['properties']['authtype']['readOnly'] = true
+  #   super.merge('Authentication' => authentication)
+  # end
```
lindgrenj6, line 39: I'm curious your opinion on this @gmcculloug, should we just `git revert 23677837939866329a48048044330ef42cb972f0` (the prior commit) for now until we can implement it in the openapi.json?

Personally I'd rather implement it there rather than having a validation for it.

app/models/authentication.rb
```diff
@@ -11,4 +11,12 @@ class Authentication < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
+
+  validate :authtype_not_updated
+
+  def authtype_not_updated
+    if !new_record? && authtype_changed?
+      errors.add(:authtype, "cannot be updated")
```
gmcculloug, line 19: @slemrmartin You can add `on: :update` to the `validate` line then the condition here is very simple:

```ruby
validate :authtype_not_updated, :on => :update

def authtype_not_updated
  errors.add(:authtype, "cannot be updated") if authtype_changed?
end
```

lib/tasks/openapi_generate.rake
gmcculloug: @lindgrenj6 I'm really not sure what is the best approach here.  As we briefly chatted about it could also be handled at the controller layer during the update call as well.

There seem to be 3 independent solutions:
1. Update OpenAPI spec to use a different object for Patch calls that define authtype as read-only.  Requires work on auto-generator
2. Model validation (this PR)
3. Limit parameters during update on the controller.

@lindgrenj6 What is the concern about having the validation?

@mkanoor Could use your thoughts on this.

lindgrenj6: I don't really have a concern with the validation itself, I would just like to solve this in the OpenAPI spec which as you said requires work on the auto-generator. 

Maybe we should just make a git-issue or jira ticket that we should enhance the auto-generator to handle different objects for POST requests, that way we can track it.

app/models/authentication.rb
slemrmartin: @gmcculloug I was thinking about it, on one side, it makes the "validation" line more obvious, on the other side, it makes the validation method less obvious so current version seems easier to understand to me (it's more important when the class becomes bigger)

lib/tasks/openapi_generate.rake
syncrou: @lindgrenj6 with https://projects.engineering.redhat.com/browse/TPINVTRY-1004 as our way to solve the issues found using the auto-generator. Do you feel the concerns for now are mitigated?

slemrmartin:
@lindgrenj6 do you thing this PR is still needed?

lindgrenj6:
@slemrmartin oh wow I forgot about this one. This will get taken care of when we mangle the OpenAPI json schema to not be auto-generated anymore. Closing.

PR #245, opened 2020-07-13
Set the first endpoint created on a source as default

lindgrenj6:
https://projects.engineering.redhat.com/browse/TPINVTRY-999

We are running into various issues where users create source/endpoint/authentication via API and if they do not pass in `default => true` in the POST /endpoints call, the source then doesn't know about an endpoint, which causes availability checks to show "Partially Available".

This way there is always a default endpoint for availability checking, even when created via the api.

### DIFF
--- a/app/models/endpoint.rb
+++ b/app/models/endpoint.rb
@@ -11,7 +11,13 @@ class Endpoint < ApplicationRecord
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
 
+  before_save :set_default, :if => proc { source.endpoints.count.zero? }
+
   def base_url_path
     URI::Generic.build(:scheme => scheme, :host => host, :port => port, :path => path).to_s
+  end
+
+  def set_default
+    self.default = true
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/endpoint.rb
```diff
@@ -11,7 +11,13 @@ class Endpoint < ApplicationRecord
...
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
 
+  before_save :set_default, :if => proc { source.endpoints.count.zero? }
+
   def base_url_path
     URI::Generic.build(:scheme => scheme, :host => host, :port => port, :path => path).to_s
   end
+
+  def set_default
+    self.default = true
```
syncrou, line 21: If they do send in `self.default = false` would this over write that?

lindgrenj6: Correct - this will make sure that the first endpoint added to a source will be marked as default which seems like a pretty common use-case when creating source/endpoint/authentication via the api.

syncrou: This really looks fine, but having a test proving your `EDIT` comment would be great. Other than that this is ready to go.

lindgrenj6: @syncrou good idea, added a spec that makes sure a delete/re-add of endpoint marks it as default again.

PR #246, opened 2020-07-27
Modified tests to use FactoryBot.

roliveri:
Converted tests to use FactoryBot.

Some more work can probably be done to clean up the TenantIdentity mixin, or eliminate the need for it completely.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ group :development, :test do
 end
 
 group :test do
-	gem "factory_bot_rails"
+  gem "factory_bot_rails"
   gem 'rspec-rails', '~>3.8'
   gem 'webmock'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ group :development, :test do
 end
 
 group :test do
+	gem "factory_bot_rails"
   gem 'rspec-rails', '~>3.8'
   gem 'webmock'
-  gem "factory_bot_rails"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -25,4 +25,5 @@ end
 group :test do
   gem 'rspec-rails', '~>3.8'
   gem 'webmock'
+  gem "factory_bot_rails"
 end
--- a/spec/controllers/api/v1x0/mixins/destroy_mixin_spec.rb
+++ b/spec/controllers/api/v1x0/mixins/destroy_mixin_spec.rb
@@ -5,8 +5,7 @@ describe Api::V1::Mixins::DestroyMixin do
     include ::Spec::Support::TenantIdentity
 
     let(:headers) {   {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
-    let!(:source_1)   { Source.create!(:source_type => source_type, :tenant => tenant, :name => "test_source 1", :uid => SecureRandom.uuid) }
-    let(:source_type) { SourceType.create!(:name => "openshift", :product_name => "OpenShift", :vendor => "Red Hat") }
+    let!(:source_1)   { create(:source, name: "test_source 1") }
     let(:client)      { instance_double("ManageIQ::Messaging::Client") }
     before do
       allow(client).to receive(:publish_topic)
@@ -21,7 +20,7 @@ describe Api::V1::Mixins::DestroyMixin do
     end
 
     context "Sub-collection:" do
-      let!(:endpoint_1) { Endpoint.create!(:role => "a", :source => source_1, :tenant => tenant) }
+      let!(:endpoint_1) { create(:endpoint, role: "a", :source => source_1) }
 
       it "delete /sources/:id/endpoint/:id fails with 404" do
         delete(api_v1x0_source_endpoints_url(source_1.id) + "/#{endpoint_1.id}", :headers => headers)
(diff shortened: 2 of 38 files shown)


roliveri:
@lindgrenj6 @slemrmartin Please review.

Gemfile
```diff
@@ -23,6 +23,7 @@ group :development, :test do
 end
 
 group :test do
+	gem "factory_bot_rails"
```
slemrmartin, line 26: please fix the indentation

spec/controllers/api/v1x0/mixins/destroy_mixin_spec.rb
```diff
@@ -21,7 +20,7 @@
     end
 
     context "Sub-collection:" do
-      let!(:endpoint_1) { Endpoint.create!(:role => "a", :source => source_1, :tenant => tenant) }
+      let!(:endpoint_1) { create(:endpoint, role: "a", :source => source_1) }
```
slemrmartin, line 23: there is mixed syntax `a:b` and `:a => b`.
Afaik this style: `a: b` is not "approved" by rubocop (the same comment for all the similar changes in this PR)

roliveri: I understand, but this leads me to ask the following:

Do we have a project-specific style and rubocop configuration file? What's "approved" or not should be based on that configuration, not the rubocop defaults - unless that configuration accepts those defaults. 

Do we, can we , should we, run rubocop on new PRs and commits like we do in MIQ?

The rubocop configuration file should be made available to developers, maybe in the guides repo, so they can check their code before committing.

If we already have what I mentioned above, then I apologize for missing it. Otherwise, I think it would be useful to implement it.

I personally like the `a: b` notation, but if it's contrary to the officially adopted project style, I would change it.

slemrmartin: we have shared rubocop rules with MIQ which defines `:a => b` style. (https://github.com/RedHatInsights/sources-api/blob/master/.rubocop.yml)

The miq-bot and rubocop check should be working, we have to look why it's not working in this repo.

lindgrenj6: @eclarizio can we get dippy-bot looking at this repo :eyes:

roliveri: @slemrmartin Thanks. Given that's the case, I the code should be changed to adhere to it.
Since I opened up the can of worms, I should put all the worms back in the can.

spec/models/application_spec.rb
```diff
@@ -1,33 +1,19 @@
...
-
-    let(:application_type)  { ApplicationType.create!("name" => "my-application", :supported_source_types => ["my-source-type"]) }
-
     subject do
-      Application.create!({
-        :application_type_id => application_type.id,
-        :source_id           => source.id,
-        :tenant              => tenant
-      })
+      create(:application, source: source)
```
syncrou, line 4: I really like all this red!

spec/requests/api/v1.0/endpoints_spec.rb
```diff
@@ -91,7 +91,7 @@ def instance_path(id)
 
     context "get" do
       it "success: with a valid id" do
-        instance = Endpoint.create!(payload.merge(:tenant => tenant))
+        instance = create(:endpoint, tenant: tenant)
```
syncrou, line 94: Is the separate tenant argument needed in these specs?

spec/requests/api/v2.0/applications_spec.rb
```diff
@@ -9,9 +9,8 @@
 
   let(:headers)          { {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
   let(:collection_path)  { "/api/v2.0/applications" }
-  let(:source)           { Source.create!(:source_type => source_type, :tenant => tenant, :uid => SecureRandom.uuid, :name => "my-source") }
-  let(:source_type)      { SourceType.create!(:name => "my-source-type", :product_name => "My Source Type", :vendor => "ACME") }
-  let(:application_type) { ApplicationType.create!("name" => "my-application", :supported_source_types => ["my-source-type"]) }
+  let(:source)           { create(:source, tenant: tenant, uid: SecureRandom.uuid) }
```
syncrou, line 12: Same, about tenant here.

spec/requests/api/v1.0/endpoints_spec.rb
roliveri: I kept the explicit, externally defined tenant (not defined by Factorybot) when the tenant needed to be accessed in an `expect`. In this case:
```
expect(response).to have_attributes(
          :status => 200,
          :parsed_body => a_hash_including("id" => instance.id.to_s, "tenant" => tenant.external_tenant)
        )
```

spec/requests/api/v2.0/applications_spec.rb
roliveri: Same here. However, it's probably not needed for every context. I just used the explicit tenant everywhere to be safe.

PR #248, opened 2020-07-28
update Dockerfile to use ubi:8.2-343

lindgrenj6:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi
+FROM registry.access.redhat.com/ubi8/ubi:8.2-343
```
syncrou, line 1: @lindgrenj6 is this the latest?

gmcculloug: What was being used before when we did not specify a version?

lindgrenj6: well it defaults to `:latest` during the build, and after a quick inspection: 
```
[ ⚛ ] crc $ docker pull registry.access.redhat.com/ubi8/ubi:latest
latest: Pulling from ubi8/ubi
Digest: sha256:8c346bf67f1f6baecf6752933a0cb3c9f3885d0eaf715639a514e4f5f0a7fe94
Status: Downloaded newer image for registry.access.redhat.com/ubi8/ubi:latest
registry.access.redhat.com/ubi8/ubi:latest
[ ⚛ ] crc $ docker pull registry.access.redhat.com/ubi8/ubi:8.2-343
8.2-343: Pulling from ubi8/ubi
Digest: sha256:8c346bf67f1f6baecf6752933a0cb3c9f3885d0eaf715639a514e4f5f0a7fe94
Status: Image is up to date for registry.access.redhat.com/ubi8/ubi:8.2-343
registry.access.redhat.com/ubi8/ubi:8.2-343
```

The SHA's line up, so we were automatically updating every build.

---

So this brings up an important question, should we switch to a tag-based model (like this) so the manifest accurately represents which tag we're running, or do we want the rolling security fixes?

syncrou: I would like to use tags representative of versions. It gives us something to point at specifically ( especially since we'll be working on automating manifest builds, ) and gives us a direct upgrade path when a CVE is found.

PR #249, opened 2020-08-10
Retry in 30 seconds if remote Kafka cluster is not connectable

bzwei:
A problem was exposed at travis build where the Kafka listener is automatically start in a separate thread. Because no Kafka host is configured, the thread restarts with an infinite loop. The enhancement is to stop the listener if local Kafka is not connectable. Another practical issue is the remote Kafka may be temporarily down. It makes sense to wait a while before retry.

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -27,6 +27,12 @@ class AvailabilityStatusListener
         process_event(event)
       end
     end
+  rescue Kafka::ConnectionError => e
+    Rails.logger.error("Cannot connect to Kafka cluster #{messaging_client_options[:host]}")
+    unless messaging_client_options[:host] == 'localhost'
+      sleep 30 # Remote Kafka may be down. Try again later
+      retry
+    end
   rescue => e
     Rails.logger.error(["Something is wrong with Kafka client: ", e.message, *e.backtrace].join($RS))
     retry
--- a/spec/lib/availability_status_listener_spec.rb
+++ b/spec/lib/availability_status_listener_spec.rb
@@ -5,14 +5,17 @@ RSpec.describe AvailabilityStatusListener do
   let(:status)     { "unavailable" }
   let(:reason)     { "host unreachable" }
   let(:now)        { Time.new(2020).utc }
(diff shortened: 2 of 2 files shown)


lib/availability_status_listener.rb
```diff
@@ -27,6 +27,12 @@ def subscribe_to_availability_status
         process_event(event)
       end
     end
+  rescue Kafka::ConnectionError => e
+    Rails.logger.error("Cannot connect to Kafka cluster #{messaging_client_options[:host]}")
+    unless messaging_client_options[:host] == 'localhost'
```
roliveri, line 32: So, is the reason we don't retry if the host is 'localhost' because we assume it's a test scenario?
Also, do the test changes cover this section of code?

bzwei: Yes. It is intended to stop retrying if it is a local test.
I added a test to cover this code.

lindgrenj6:
@bzwei is this change needed for rdkafka? At the very least we should probably change the exception class to match up with whatever the rdkakfa connection exception is. 

I like the change so if we could get it updated that would be great!

lib/availability_status_listener.rb
```diff
@@ -27,6 +27,12 @@ def subscribe_to_availability_status
         process_event(event)
       end
     end
+  rescue Kafka::ConnectionError => e
```
slemrmartin, line 30: Regarding my fixes in the https://github.com/RedHatInsights/topological_inventory-ingress_api/pull/102 and https://github.com/RedHatInsights/topological_inventory-ingress_api/pull/103 this class will be unknown with the rdkafka solution.

spec/lib/availability_status_listener_spec.rb
```diff
@@ -101,4 +104,21 @@
       end
     end
   end
+
+  context "when kafka at localhost is not available" do
```
slemrmartin, line 108: actually it's testing only the "I don't care" part from production's (stage/qa/ci) point of view

slemrmartin (changes requested):

bzwei:
The fix is no longer needed after we switch to rdkafka.

PR #255, opened 2020-09-11
Add Migration for Cost Sources to use Application Authentication resources

lindgrenj6:
This will make the availability check always fall back to the Application status, since no endpoint is available.

Tested locally and the fallback code works :+1:, now we just need the UI to create them under the application instead. This should probably go in after that UI change so we don't have to run this again.

JIRA: https://projects.engineering.redhat.com/browse/TPINVTRY-1082

\# TODO:
- [x] Cost Management updated to use AppAuth vs EndpointAuth
- [x] Subscription Watch updated to use AppAuth vs EndpointAuth
- [x] CM UI Updated to create AppAuth instead of EndpointAuth and update those as well
- [x] SW UI Updated to create AppAuth instead of EndpointAuth and update those as well

cc @syncrou @dccurtis

### DIFF
--- a/db/migrate/20200911162327_move_endpoint_authentications_to_application_authentication.rb
+++ b/db/migrate/20200911162327_move_endpoint_authentications_to_application_authentication.rb
@@ -1,0 +1,55 @@
+class MoveEndpointAuthenticationsToApplicationAuthentication < ActiveRecord::Migration[5.2]
+  def up
+    [cost, subscription_watch].each do |type|
+      # Skip the migration if there is no data to migrate.
+      next if type.nil?
+
+      ActiveRecord::Base.transaction do
+        Application.where(:application_type_id => type.id).each do |app|
+          source = app.source
+          authentications = source.endpoints.flat_map(&:authentications)
+
+          authentications.each do |auth|
+            auth.update!(:resource => app)
+
+            ApplicationAuthentication.find_or_create_by(
+              :authentication_id => auth.id,
+              :application_id    => app.id,
+              :tenant            => source.tenant
+            )
+          end
+
+          source.endpoints.map(&:delete)
+        end
+      end
+    end
+  end
+
+  def down
+    [cost, subscription_watch].each do |type|
+      next if type.nil?
+
+      ActiveRecord::Base.transaction do
+        Application.where(:application_type_id => type.id).each do |app|
+          source = app.source
+          authentications = source.applications.flat_map(&:authentications)
+
+          authentications.map(&:application_authentications).each(&:delete)
+
+          endpoint = source.endpoints.find_or_create_by(:tenant => source.tenant, :default => true)
+          authentications.each { |auth| auth.update!(:resource => endpoint) }
+        end
+      end
+    end
+  end
+
+  private
+
+  def cost
+    @cost ||= ApplicationType.find_by(:name => "/insights/platform/cost-management")
+  end
+
+  def subscription_watch
+    @subscription_watch ||= ApplicationType.find_by(:name => "/insights/platform/cloud-meter")
+  end
+end


lindgrenj6:
cc @rvsia

syncrou:
@lindgrenj6 was it expected that the specs were red?

dccurtis:
@lindgrenj6 Can you hold off on merging this.  It looks like there are some small changes I need to make on the Cost side.

lindgrenj6:
@dccurtis yep, we probably aren't going to merge this until the UI is fixed :+1:

dccurtis:
@lindgrenj6 https://github.com/project-koku/koku/pull/2398 has been merged.  Cost supports both Application and Endpoint authentications with this change.

rvsia:
@lindgrenj6 The ui is ready (https://github.com/RedHatInsights/sources-ui/pull/438) - for cost sources (aws/azure) it does not create empty endpoints anymore.

db/migrate/20200911162327_move_endpoint_authentications_to_application_authentication.rb
```diff
@@ -0,0 +1,46 @@
+class MoveEndpointAuthenticationsToApplicationAuthentication < ActiveRecord::Migration[5.2]
+  def up
+    [cost, subscription_watch].each do |type|
+      # Skip the migration if there is no data to migrate.
+      return true if type.nil?
+
+      ActiveRecord::Base.transaction do
+        Application.where(:application_type_id => type.id).each do |app|
+          source = app.source
+          authentications = source.endpoints.flat_map(&:authentications)
```
slemrmartin, line 10: @lindgrenj6 
We have also `aplication_authentications` table
model `Application`: `has_many :authentications, :through => :application_authentications`

We should review if there is use case for this relation and remove it, if not, it's confusing

slemrmartin: @rvsia is UI using this M:N relationship somewhere?

rvsia: @slemrmartin Yeah. It connects all applications with their assigned authentications.

slemrmartin: ok, it means this migration should use the applications_authentications too

rvsia: @slemrmartin I think these relationships should already exist. At least, the UI creates them since the table was introduced.

However, with the latest changes, we have got two 'connections' between an app and auths.

I am kinda discussing this issue here https://issues.redhat.com/browse/RHCLOUD-9526

werwty:
The Subscription Watch changes to use AppAuth have been merged.

lindgrenj6:
@rvsia @dccurtis @syncrou 

So it sounds like this is about ready to be merged - @rvsia I remember hearing that we can just do a :heavy_check_mark: in the config for SW UI to turn on the changes? If thats true then we're probably good to go here.

rvsia:
@lindgrenj6 Yeah, working on that now.

rvsia:
@lindgrenj6 The UI is ready! :crossed_swords:

syncrou (approved):
It seems like this is ready to go, but there are no Approvals yet.  

LGTM 👍 

@slemrmartin is this ready to go?

slemrmartin:
@syncrou @lindgrenj6 no, we should discuss my review as this is breaking intended architecture.

slemrmartin:
and read my comment https://issues.redhat.com/browse/RHCLOUD-9526 please, it's related to this and the future

lindgrenj6:
After speaking with @slemrmartin to better understand the problem I was unaware that `ApplicationAuthentication` was an actual table - I wasn't thinking that was a table due to the fact we have the polymorphic relationship on the Authentication model. 

I added some logic to dynamically create them on up, remove them on down.

slemrmartin (approved):
:+1: LGTM

lindgrenj6:
@roliveri @syncrou I think this is good to go.

PR #257, opened 2020-09-14
Remove unused source types

rvsia:
Closes https://issues.redhat.com/browse/TPINVTRY-1077

This is a draft PR for discussing how to properly solve this issue. (If it's going to be an issue at all.)

There are basically two options:

1) Remove unused source types from API (**this PR**)

2) Hide unused sources types in the UI

Personally, I would go with the first option as I am trying to make the UI as much consistent to the API as possible. Data should be controlled by the API, not in the UI that just shows them. Hiding them would introduce two additional issues: users won't see sources created from the API and there would have to be a list with hidden types (so more code to keep updated and more code with possible bugs). Also, it doesn't make much sense to have unusable types defined in the API.

What do you think @syncrou ?

cc @beccaglass

TODO:
- [x] migration
   - I am not sure how to migrate these data.

### DIFF
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -214,14 +214,6 @@ openshift:
           :is: true
   :vendor: Red Hat
   :icon_url: "/apps/frontend-assets/platform-logos/openshift-container-platform.svg"
-openstack:
-  :product_name: Red Hat OpenStack
-  :vendor: Red Hat
-  :icon_url: "/apps/chrome/assets/images/platform-icons/openstack.svg"
-ovirt:
-  :product_name: Red Hat Virtualization
-  :vendor: Red Hat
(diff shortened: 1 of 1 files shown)


gmcculloug:
@rvsia Your TODO item in the description is already handled as part of the server startup:  https://github.com/RedHatInsights/sources-api/blob/master/docker-assets/run_rails_server#L32

Is that the only thing holding this back from moving forward?

rvsia:
@gmcculloug 

I was mostly curious if the IDs of the other sources will be preserved (as I don't have too much experience with seeding in Rails apps.)

Otherwise, the only missing thing is an approve. :heavy_check_mark:

syncrou (approved):
This looks good @rvsia, I would rather move these now since we're not using them. Also will be easy to bring them back if we need them again.

I agree that I would rather the API controls this and the UI responds to what the API returns.

Nice!

PR #262, opened 2020-09-29
Updated swagger_spec.rb

roliveri:
to account for changes made in: https://github.com/RedHatInsights/insights-api-common-rails/pull/173


syncrou (approved):
LGTM 👍

PR #263, opened 2020-10-01
Upgrading the insights-api-common gem to 4.1.4 version.

pkomanek:
Upgrading the `insights-api-common` gem to 4.1.4 version that no longer allows redirects from major versions on a POST.

This PR is based on https://issues.redhat.com/browse/RHCLOUD-9585.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
 gem 'cloudwatchlogger',     '~> 0.2.1'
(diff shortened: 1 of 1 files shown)


lindgrenj6 (approved):

lindgrenj6:
We don't need this PR anymore, the `~> 4.0` gets us the new version already, it was pulled in on an image rebuild. If we want to lock it down to `~> 4.1.0` to get us 4.1.0 -> 4.1.x we can discuss that as well, but I'm not sure we need to do that. 

Validated on CI: 
```
[ master ⚛ ] crc $ k exec -it sources-api-997-cvvht -- gem list | grep insights-api-common
insights-api-common (4.1.5)
```

PR #267, opened 2020-10-14
Move to Pundit Gem for authorization

lindgrenj6:
Setting up some killer RBAC for Sources!

EDIT: Deciding to break this up so it's easier to review. 

This PR is for the first part of this ticket: https://issues.redhat.com/browse/RHCLOUD-9963

Basically all this PR is doing is moving everything over to pundit-based access checks. It is using the same `org_admin?` check as before basically but just going through pundit instead of doing it manually. 

\# TODO:
- [x] Add pundit gem
- [x] Create policies for models
- [x] Wire up policies in controllers (using old org_admin? logic)
~- [ ] Add RBAC lib class~
~- [ ] Add ACL table to cache rbac results.~

### DIFF
--- a/app/controllers/api/v1/mixins/destroy_mixin.rb
+++ b/app/controllers/api/v1/mixins/destroy_mixin.rb
@@ -3,7 +3,10 @@ module Api
     module Mixins
       module DestroyMixin
         def destroy
-          model.destroy(params.require(:id))
+          record = model.find(params.require(:id))
+          authorize(record)
+
+          record.destroy
           head :no_content
         end
       end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,8 +4,7 @@ class ApplicationController < ActionController::API
   include Insights::API::Common::ApplicationControllerMixins::ExceptionHandling
   include Insights::API::Common::ApplicationControllerMixins::RequestBodyValidation
   include Insights::API::Common::ApplicationControllerMixins::RequestPath
-
-  class RbacError < StandardError; end
+  include Pundit
 
   BLACKLIST_PARAMS = [:tenant].freeze
 
@@ -44,7 +43,6 @@ class ApplicationController < ActionController::API
     Insights::API::Common::Request.with_request(request) do |current|
       begin
         if Tenant.tenancy_enabled? && current.required_auth?
-          validate_request_allowed!(current)
           validate_request_entitled!(current)
 
           tenant = Tenant.find_or_create_by(:external_tenant => current.tenant)
@@ -58,19 +56,8 @@ class ApplicationController < ActionController::API
       rescue Insights::API::Common::EntitlementError
         error_document = Insights::API::Common::ErrorDocument.new.add('403', 'Forbidden')
         render :json => error_document.to_h, :status => error_document.status
-      rescue RbacError
-        error_document = Insights::API::Common::ErrorDocument.new.add('403', 'Forbidden due to Rbac')
-        render :json => error_document.to_h, :status => error_document.status
       end
     end
-  end
-
-  def request_is_readonly?
-    [:get, :head, :options].include?(request.request_method_symbol)
-  end
-
-  def request_is_create?
-    request.request_method_symbol == :post
   end
 
   def validate_request_entitled!(current)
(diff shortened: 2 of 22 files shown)


lindgrenj6:
Ignore the dippy bot on that error, it will go away when we move more of the custom exception mappings to the initializer hash.

app/controllers/application_controller.rb
```diff
@@ -58,43 +56,17 @@ def with_current_request
...
         error_document = Insights::API::Common::ErrorDocument.new.add('403', 'Forbidden')
         render :json => error_document.to_h, :status => error_document.status
-      rescue RbacError
-        error_document = Insights::API::Common::ErrorDocument.new.add('403', 'Forbidden due to Rbac')
-        render :json => error_document.to_h, :status => error_document.status
       end
     end
   end
 
-  def request_is_readonly?
```
slemrmartin, line 68: you can remove it also from graphql_controller if not needed

slemrmartin (changes requested):
I'm missing authorizations for:
- v1/applications_controller#create
- v1/source_types_controller#create
- v2/application_authentications_controller#create

slemrmartin:
@syncrou this is not exactly related to this PR, but if we don't want to allow SourceType#create, should we offer API v1 which allows it?

lindgrenj6:
@slemrmartin yep - I'll remove the method in this PR, we removed the route + openapi definition in https://github.com/RedHatInsights/sources-api/pull/250 when pentesting was breaking the stage env by creating bad SourceTypes.

app/controllers/api/v1/mixins/destroy_mixin.rb
```diff
@@ -3,7 +3,10 @@ module V1
     module Mixins
       module DestroyMixin
         def destroy
-          model.destroy(params.require(:id))
+          record = model.find(params.require(:id))
+          authorize(record)
```
syncrou, line 7: I'm guessing `authorize(record)` raises an exception if there is an illegal attempt? 

Otherwise was wondering what stops `record.destroy` from getting executed.

app/policies/application_authentication_policy.rb
```diff
@@ -0,0 +1,7 @@
+class ApplicationAuthenticationPolicy < DefaultPolicy
+  def create?
```
syncrou, line 2: Is there anyway to consolidate the contents of these 3 files into one via a mixin?

Just asking since the `create?`, and aliases methods are all duplicated.

app/policies/endpoint_policy.rb
```diff
@@ -0,0 +1,7 @@
+class EndpointPolicy < DefaultPolicy
+  def create?
```
syncrou, line 2: These to I guess. Since all the method definitions are the exact same.

app/policies/source_policy.rb
```diff
@@ -0,0 +1,7 @@
+class SourcePolicy < DefaultPolicy
+  def create?
```
syncrou, line 2: Same

app/controllers/api/v1/mixins/destroy_mixin.rb
lindgrenj6: Yep - exactly. Pundit raises `Pundit::NotAuthorized` when the authorization fails.

app/policies/application_authentication_policy.rb
lindgrenj6: Oh I like that idea a lot. Implemented!

config/initializers/custom_exception_mappings.rb
```diff
@@ -0,0 +1,6 @@
+ActionDispatch::ExceptionWrapper.rescue_responses.merge!(
+  "Pundit::NotAuthorizedError"              => :forbidden,
```
slemrmartin, line 2: Does it return `Insights::API::Common::ErrorDocument` like RbacError returned?

lindgrenj6: Yep, that is handled here: https://github.com/RedHatInsights/insights-api-common-rails/blob/master/lib/insights/api/common/application_controller_mixins/exception_handling.rb

which is mixed in the base ApplicationController.

slemrmartin (approved):
:+1: LGTM, maybe small naming thing. 
I would like more `WritePolicyMixin` than `WriteCheckMixin`

lindgrenj6:
>  I would like more WritePolicyMixin than WriteCheckMixin

Oh yeah I like that too, better describes the mixin type. Updated!

roliveri:
Just to confirm. I'm assuming the behavior hasn't changed, so the existing tests cover these changes. Is that correct?

lindgrenj6:
@roliveri yes, that is correct. The existing `org_admin?` tests just go through Pundit now rather than the (now removed) checks in `app/controllers/application_controller.rb`.

roliveri (approved):

PR #268, opened 2020-10-22
Add call out to RBAC for write-access check

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-9964

- [x] Built off of #267 so that one needs to go in first.

Pretty self-explanatory, this just adds in the call out to the RBAC service for authorization when doing anything that requires write access (admin-only for now). 

The only permission as of right now is `sources:*:*` which we will potentially expand later.

### DIFF
--- a/app/policies/default_policy.rb
+++ b/app/policies/default_policy.rb
@@ -30,10 +30,12 @@ class DefaultPolicy
   alias delete? destroy?
 
   def admin?
-    return true unless Insights::API::Common::RBAC::Access.enabled?
+    return true unless Sources::RBAC::Access.enabled?
 
-    user.system.present? || user.user&.org_admin?
+    user.system.present? || user.user&.org_admin? || write_access?
   end
+
+  delegate :write_access?, :to => Sources::RBAC::Access
 
   class Scope
     attr_reader :user, :scope
--- a/lib/sources/rbac/access.rb
+++ b/lib/sources/rbac/access.rb
@@ -1,0 +1,19 @@
+module Sources
+  module RBAC
+    class Access
+      class << self
+        def enabled?
+          Insights::API::Common::RBAC::Access.enabled?
+        end
+
+        def write_access?
+          access.accessible?("*", "*") # sources:*:* is the only permission as of now.
+        end
+
+        def access
+          @access ||= Insights::API::Common::RBAC::Access.new.process
+        end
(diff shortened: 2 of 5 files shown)


roliveri:
Do you have a handle on the spec failures?

lindgrenj6:
@roliveri Looked like they were rdkafka failures, lets see if opening/closing it works.

lindgrenj6:
@roliveri there we go, sorry about that. I had one that I hadn't committed yet.

.rubocop.yml
```diff
@@ -1,3 +1,3 @@
 inherit_from:
-- https://raw.githubusercontent.com/ManageIQ/guides/master/.rubocop_base.yml
+- https://raw.githubusercontent.com/RedHatInsights/insights-api-common-rails/master/.rubocop_base.yml
```
roliveri, line 2: @lindgrenj6  Is this change the cause of the Code Climate failure?

lindgrenj6: Not because of the code climate - it was for dippy-bot, that way it can run the rubocop rules successfully. Codeclimate won't quit yelling at us over the copied code of mixing-in various controllers.

roliveri: But the failure doesn't seem to be due to copied code. It's failing in setup:
```
codeclimate prepare
Failed fetching https://raw.githubusercontent.com/ManageIQ/guides/master/.rubocop_base.yml: code=404 body=404: Not Found
```
That's why I ask about that specific change.

lindgrenj6: Oh - yeah codeclimate.yml will need to be updated too. I actually think I'll pull this change out and open a new PR with both of those changes in it.

slemrmartin: I'm also for separated PR

slemrmartin (changes requested):
LGTM, I'm just for moving the rubocop change to insigths api common (together with next line, which depends on https://github.com/RedHatInsights/insights-api-common-rails/pull/211

lindgrenj6:
@slemrmartin @roliveri pulled and opened here: https://github.com/RedHatInsights/sources-api/pull/272

slemrmartin (approved):
:+1: LGTM

lindgrenj6:
@roliveri this is ready to go!

PR #269, opened 2020-10-27
Publishing Kafka event ApplicationAuthentication.destroy

slemrmartin:
After destroying of Application or Authentication, if ApplicationAuthentication relation exists, kafka event `ApplicationAuthentication.destroy` has to be raised as a cascade delete operation

---

[RHCLOUD-10263](https://issues.redhat.com/browse/RHCLOUD-10263)

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -5,7 +5,7 @@ class Application < ApplicationRecord
   belongs_to :source
   belongs_to :application_type
 
-  has_many :application_authentications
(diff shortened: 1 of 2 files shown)


lindgrenj6 (approved):

syncrou (commented):
@slemrmartin, @rvsia if we cascade delete, is there potential that shared authentications will be removed?

syncrou (approved):
If we do not destroy the authentications than I am fine with this. LGTM 👍

PR #273, opened 2020-10-29
Memoize `Sources::Api::Messaging.client` instead of using thread-local storage

lindgrenj6:
Same change as on topo + ingress, memoizing the api client so we don't instantiate (and then never close) a bunch of messaging clients.


slemrmartin (approved):
:+1:

PR #275, opened 2020-11-16
Remove check for whether a topic exists when checking Source#availability_status

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-10747

It seems we are running into this issue a lot where the message producers will "forget" that a topic exists, and proceed to not produce the availability status messages anymore, seemingly happens in random intervals but the logs are filled with messages like this:
```json
{
  "@timestamp": "2020-11-16T16:04:31.059618 ",
  "hostname": "sources-api-774b8d79cc-6m2zs",
  "pid": 32,
  "tid": "2afa089bbc0c",
  "level": "err",
  "message": "Not publishing message to non-existing topic: Source#availability_check [{\"source_id\"=>1307, \"topic\"=>\"platform.topological-inventory.operations-azure\"}]",
  "request_id": "a240ffd873bc45fea2e6665e8b925552"
}
```

even though the topic definitely exists.

---

This PR is one possible solution - rather than checking if a topic exists and not producing, lets just try and produce every time and just rescue/log. 
Open to other suggestions as well, another solution would be to potentially instantiate a new producer every time and just close it properly? But I'm not sure what the cost would be as far as resources go for that since opening up a FFI connection isn't cheap.

### DIFF
--- a/app/controllers/api/v1/sources_controller.rb
+++ b/app/controllers/api/v1/sources_controller.rb
@@ -26,7 +26,7 @@ module Api
 
         logger.info("Initiating Source#availability_check [#{{"source_id" => source.id, "topic" => topic}}]")
 
-        if Sources::Api::Messaging.topics.include?(topic)
+        begin
           logger.debug("Publishing message for Source#availability_check [#{{"source_id" => source.id, "topic" => topic}}]")
 
           Sources::Api::Messaging.client.publish_topic(
@@ -43,8 +43,8 @@ module Api
           )
 
           logger.debug("Publishing message for Source#availability_check [#{{"source_id" => source.id, "topic" => topic}}]...Complete")
-        else
-          logger.error("Not publishing message to non-existing topic: Source#availability_check [#{{"source_id" => source.id, "topic" => topic}}]")
+        rescue e
+          logger.error("Hit error attempting to publish [#{{"source_id" => source.id, "topic" => topic}}] during Source#availability_check: #{e.message}")
         end
 
         check_application_availability(source)
--- a/lib/sources/api/messaging.rb
+++ b/lib/sources/api/messaging.rb
@@ -13,10 +13,6 @@ module Sources
           :encoding => "json"
         )
       end
-
-      cache_with_timeout(:topics) do
-        client.topics || []
-      end
     end
   end
 end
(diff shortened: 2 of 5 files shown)


app/controllers/api/v1/sources_controller.rb
```diff
@@ -26,7 +26,7 @@ def check_availability
 
         logger.info("Initiating Source#availability_check [#{{"source_id" => source.id, "topic" => topic}}]")
 
-        if Sources::Api::Messaging.topics.include?(topic)
```
syncrou, line 29: Is this behavior than just tech debt that no one is using? ie should it get ✂️ too?

lindgrenj6: Yeah afaik this is the only place we were using the topics method, so it can probably be removed from the manageiq-messaging gem.

slemrmartin:
@lindgrenj6 we have to check behaviour on stage/prod carefully. This check was added for ruby-kafka implementation because it (==kafka client?) accumulated undelivered messages to non-existent topics and then blocked delivering messages to all topics (and it was difficult to detect this behaviour).
Is rdkafka implementation without this issue?

lindgrenj6:
@slemrmartin interesting, I was wondering why this check was here in the first place. That problem may/may not exist in the rdkafka implementation, but the other side of the coin is we _know_ these topics exist in stage/prod always at the very least. 

Since they're statically defined in the `platform-mq` repository and created by the strimzi operator  there should _never_ be a time when it fails to produce messages like before if that was happening. 

So I think it should be pretty safe to remove the check as long as we have all of the operations topics created everywhere (which I think we do).

lindgrenj6:
@slemrmartin @roliveri can we get some motion on this? happening again on production when Catalog is trying to add a source.

syncrou (approved):
Per this comment: https://github.com/RedHatInsights/sources-api/pull/275#issuecomment-732335963 - I think we should be okay to move ahead on this. We're sort of stuck now anyway and doing nothing isn't helping fix the problem.  

@roliveri can I get some 👀 on this as well?

roliveri (approved):
While I don't know enough to understand how/why message producers can forget that a topic exists, this seems to be a good remedy for that situation.

The only concern I have is that now, failures not due to missing topics will be handled differently, in that they will also be handled by the newly introduced rescue. Not sure if this is an issue, but it is a possible difference in behavior.

PR #278, opened 2020-11-20
Add `source_id` to Authentications and Source has_many relationship

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-10074

Adding a relationship between authentications and sources:

Source has_many Authentications

EDIT: also added `/sources/:id/authentications` endpoint per request from @rvsia

### DIFF
--- a/app/models/authentication.rb
+++ b/app/models/authentication.rb
@@ -5,10 +5,19 @@ class Authentication < ApplicationRecord
   encrypt_column :password
 
   belongs_to :resource, :polymorphic => true
+  belongs_to :source, :optional => true
 
   has_many :application_authentications, :dependent => :destroy
   has_many :applications, :through => :application_authentications
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
+
+  before_validation :set_source
+
+  private
+
+  def set_source
+    self.source_id = resource&.source_id
+  end
 end
--- a/db/migrate/20201120163548_add_source_to_authentications.rb
+++ b/db/migrate/20201120163548_add_source_to_authentications.rb
@@ -1,0 +1,17 @@
+class AddSourceToAuthentications < ActiveRecord::Migration[5.2]
+  def up
+    add_column :authentications, :source_id, :bigint
+    Authentication.transaction do
+      Authentication.find_each do |auth|
+        # apparently we have some dangling endpoints/applications, so we need to skip them.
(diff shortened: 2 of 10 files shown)


app/models/authentication.rb
```diff
@@ -5,10 +5,19 @@ class Authentication < ApplicationRecord
...
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available unavailable] }, :allow_nil => true
+
+  before_validation :set_source
+
+  private
+
+  def set_source
+    self.source_id = resource.source_id
```
slemrmartin, line 21: resource can be nil

db/migrate/20201120163548_add_source_to_authentications.rb
```diff
@@ -0,0 +1,17 @@
+class AddSourceToAuthentications < ActiveRecord::Migration[5.2]
+  def up
+    add_column :authentications, :source_id, :bigint
+    Authentication.transaction do
+      Authentication.all.each do |auth|
```
slemrmartin, line 5: Please use `find_each`

public/doc/openapi-3-v1.0.json
```diff
@@ -1386,6 +1386,9 @@
             "example": "Endpoint",
             "type": "string"
           },
+          "source_id": {
```
slemrmartin, line 1389: should we update v1.0?

slemrmartin (changes requested):
looks good, only few minor things

public/doc/openapi-3-v1.0.json
lindgrenj6: Not sure, I thought we were supposed to but should we just only update the latest?

app/models/authentication.rb
lindgrenj6: Well, the fun thing is that it won't allow it to be created unless the source exists, so `nil` won't work. I'll add `:optional => true` to the `belongs_to` statement to allow a blank source_id.

public/doc/openapi-3-v1.0.json
slemrmartin: I'm not sure too, theoretically yes, but practically v1.0 is used by someone so let's keep it updated

slemrmartin (approved):
:+1: LGTM, please generate sources-api-client for this PR

lindgrenj6:
@slemrmartin https://github.com/RedHatInsights/sources-api-client-ruby/pull/24

PR #283, opened 2020-12-02
Remove memoization of rbac access class variable

lindgrenj6:
I think this is the problem we're seeing, since it was a class var and it was memoized the first person who logged in sets the permissions for the rest of the hits. Not ideal.


syncrou (approved):
`class << self` for the win...  Nice find. LGTM 👍

PR #285, opened 2020-12-07
Kafka availability_status_listener fix.

pkomanek:
Kafka availability_status_listener fix.

Based on: https://github.com/RedHatInsights/sources-api/issues/286
Depends on: [ ] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/68

In the pictures bellow you can see the sources-api command line output containing DB update for all realted tables. 

services/ansible-tower-operations.sh :
Screenshot from 2020-12-07 17-11-41
sources-api:
Screenshot from 2020-12-07 17-11-11
DB:
Screenshot from 2020-12-07 17-12-07

EDIT: 
 - Add missing message type for logging message. 
    - example: Kafka message **availability_status** received with payload....

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -38,25 +38,26 @@ class AvailabilityStatusListener
     Rails.logger.info("Kafka message #{event.message} received with payload: #{event.payload}")
     return unless event.message == EVENT_AVAILABILITY_STATUS
 
-    model_class = event.payload["resource_type"].classify.constantize
+    payload = JSON.parse(event.payload)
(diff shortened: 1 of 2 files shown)


slemrmartin (changes requested):
@pkomanek I found the root cause, all other services are using `client.publish_topic`, but operations are using `client.publish_message`. 
Although it's similar in Kafka (except there differences), I would rather unify the **publish_** calls - so to change it in the ProvidersCommon::Source (as publish_message is intended for one-to-one communication which is not really supported in Kafka)

slemrmartin (approved):
:+1: LGTM

PR #289, opened 2020-12-15
[WIP]Graphql provide metadata

pkomanek:
Add metadata for GraphQL requests. It includes the count, which ignores limits and offsets for all root elements. This behaviour is similar to our REST-API and it makes both of them to be more consistent.

Data output:
 * query: 
```
query { 
  applications(offset:1, limit:2) {
    application_type_id,
    id
  },
  endpoints(offset:2, limit:1) {
    id,
    scheme,
    host,
    port,
    path
  } 
}
```
 * before:
```
{
  "data":{
    "applications":[
      {"application_type_id":"4","id":"2"},
      {"application_type_id":"4","id":"3"}
    ],
    "endpoints":[{"id":"3","scheme":"https","host":"tower.example.com","port":null,"path":"/"}]
  }
}
```
 * after: 
```
{
  "data":{
    "applications":[
      {"application_type_id":"4","id":"2"},
      {"application_type_id":"4","id":"3"}
    ],
    "endpoints":[{"id":"3","scheme":"https","host":"tower.example.com","port":null,"path":"/"}]
  },
  "meta":{"count":{"applications":4,"endpoints":5}}
}
```

This PR is based on https://issues.redhat.com/browse/RHCLOUD-9332

### DIFF
--- a/app/controllers/api/v1/graphql_controller.rb
+++ b/app/controllers/api/v1/graphql_controller.rb
@@ -3,6 +3,10 @@ require "insights/api/common/graphql"
 module Api
   module V1
     class GraphqlController < ApplicationController
+      class << self
+        attr_accessor :limits_and_offsets
+      end
+
       def query
         schema_overlay = {
           "^.+$" => {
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/graphql_controller.rb
```diff
@@ -19,7 +23,86 @@ def query
...
+      end
+
+      def apply_limit(data, limit)
+        data[0..limit - 1]
+      end
+    end
+  end
+end
+
+module GraphQL
```
slemrmartin, line 71: I expect the Monkey patching will be there only during WIP, right?

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -19,7 +23,86 @@ def query
...
+  module Execution
+    class Execute
+      class << self
+        def begin_query(query, _multiplex)
+          remove_offsets_and_limits(query)
```
slemrmartin, line 76: I think it can't work this way, removing limit from request can cause returning milions of records. SQL count should be used instead

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -19,7 +23,86 @@ def query
...
+          query.arguments.delete_if do |arg|
+            if arg.name == 'offset' || arg.name == 'limit'
+              limits_and_offsets[arg.name] = arg.value
+              true
+            else
+              false
+            end
+          end
+
+          Api::V1::GraphqlController.limits_and_offsets ||= {}
```
slemrmartin, line 103: This is not thread-safe and api-version safe

app/controllers/api/v1/graphql_controller.rb
```diff
@@ -3,6 +3,10 @@
 module Api
   module V1
     class GraphqlController < ApplicationController
+      class << self
+        attr_accessor :limits_and_offsets
```
slemrmartin, line 7: I'm not sure if this will work with more parallel requests

slemrmartin (changes requested):

PR #292, opened 2020-12-17
Add availability_status concern

pkomanek:
Since we cannot be sure about the `availability_status` after attribute update, we are adding the availability_status concern, which removes `availability_status`, `availability_status_error` and `last_checked_at` attributes from `Sources`, `Applications`, `Endpoints` and `Authentications`.

This PR is based on https://issues.redhat.com/browse/RHCLOUD-10278.

### DIFF
--- a/app/models/authentication.rb
+++ b/app/models/authentication.rb
@@ -2,6 +2,7 @@ class Authentication < ApplicationRecord
   include PasswordConcern
   include TenancyConcern
   include EventConcern
+  include AvailabilityStatusConcern
   encrypt_column :password
 
   belongs_to :resource, :polymorphic => true
--- a/app/models/concerns/availability_status_concern.rb
+++ b/app/models/concerns/availability_status_concern.rb
@@ -1,0 +1,31 @@
+module AvailabilityStatusConcern
+  extend ActiveSupport::Concern
+
+  included do
+    before_update :update_status
+  end
+
+  private
+
+  IGNORE_LIST = [
+    "availability_status",
+    "availability_status_error",
+    "last_available_at",
+    "last_checked_at",
+    "updated_at",
+    "name"
+  ].freeze
+
+  def update_status
+    updated_attributes = changed - IGNORE_LIST
+
+    if updated_attributes.any?
+      self.availability_status = nil
+      self.last_checked_at = nil
+
+      if respond_to?(:availability_status_error)
+        self.availability_status_error = nil
+      end
+    end
+  end
+end
--- a/spec/models/shared/availability_status.rb
+++ b/spec/models/shared/availability_status.rb
@@ -1,0 +1,80 @@
+RSpec.shared_examples "availability_status" do
+  # record, update and no_update variables come from caller spec
+  describe "#before_update" do
+    let(:timestamp)                 { Time.current }
+    let(:new_timestamp)             { Time.current + 1.hour }
+    let(:available_status)          { "available" }
+    let(:unavailable_status)        { "unavailable" }
+    let(:new_name)                  { "new_name" }
+    let(:availability_status_error) { "availability_status_error" }
+
+    let(:ignored_attributes) do
+      new_attributes = {
+        :availability_status => unavailable_status,
+        :last_checked_at     => new_timestamp,
+        :last_available_at   => new_timestamp
+      }
+
+      if record.respond_to?(:updated_at)
+        new_attributes[:updated_at] = new_timestamp
+      end
+
+      if record.respond_to?(:availability_status_error)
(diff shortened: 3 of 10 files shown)


app/models/authentication.rb
```diff
@@ -2,6 +2,7 @@ class Authentication < ApplicationRecord
   include PasswordConcern
   include TenancyConcern
   include EventConcern
+  include AvailabilityStatusConcern
```
slemrmartin, line 5: Authentication doesn't have availability status, can be removed

app/models/concerns/availability_status_concern.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  STATUS_RELATED_ATTRS = [
+    "availability_status",
+    "last_checked_at",
+    "updated_at",
+    "last_available_at",
+    "availability_status_error"
+  ].freeze
+
+  def update_status
```
slemrmartin, line 18: I'm wondering whether the availability status update doesn't cause availability status reset by this code?

app/models/concerns/availability_status_concern.rb
```diff
@@ -0,0 +1,25 @@
...
+    updated_attributes = self.changed - STATUS_RELATED_ATTRS
+    unless updated_attributes.empty?
+      self.availability_status = nil
+      self.last_checked_at = nil
```
slemrmartin, line 22: availability status error can be also cleared

app/models/authentication.rb
pkomanek: Yes it does. You can find it in the database scheme bellow. I am not sure, if we are using these fields, since I have them empty in all of my data. If we don't, we should remove these DB columns. What do you thin about that @slemrmartin and @lindgrenj6 ?
Screenshot from 2020-12-18 12-16-21

spec/models/shared/availability_status.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    before do
+      record.availability_status = available_status
+      record.last_checked_at = timestamp
+    end
+
+    context "with changes" do
+      it "sets availability_status to nil" do
+        record.update!({:id=>record.id + 1})
```
lindgrenj6, line 40: Let's change this to `:name` just to be more realistic. We would never actually update the ID.

spec/models/shared/availability_status.rb
```diff
@@ -0,0 +1,80 @@
...
+
+        if record.respond_to?(:name)
+          expect(record.name).to eq(new_name)
+        end
+      end
+    end
+
+    context "without changes" do
+      it "keeps availability_status unchanged" do
+        record.update!({:id=>record.id})
```
lindgrenj6, line 73: Same here, change to `:name`.

lindgrenj6 (commented):
Just a few minor things, looks good otherwise.

One thing we could probably discuss eventually would be to also queue an availability check after an edit, that way its immediately checked. I'm not sure if the UI does that already.

spec/models/shared/availability_status.rb
pkomanek: @lindgrenj6 I was talking with Martin before lockdown and we agreed that the `name` attribute change should not clear the availability_status and we putted him into the [ignore list](https://github.com/RedHatInsights/sources-api/blob/c968069fe5f6f3bf2278d1c8b6ba378dbb418b49/app/models/concerns/availability_status_concern.rb#L10-L17). Also I was checking the database and the only attributes that comes out from the union of all related models(Endpoints, Applications, Sources and Authentications) are `id` and `tenant_id`. Since this is shared example group, I think we should either keep with one of them or make specific change for every model similar to the record forwarding [here](https://github.com/RedHatInsights/sources-api/blob/c968069fe5f6f3bf2278d1c8b6ba378dbb418b49/spec/models/authentication_spec.rb#L5). What do you think Jacob?

lindgrenj6: @pkomanek well it doesn't have to be name, that makes sense. lets just update something in the test that isn't `:id` since updating id will never happen in real life. `availability_status` would work.

syncrou (commented):
This is looking good to me. Is it ready to go out @slemrmartin @lindgrenj6

app/models/authentication.rb
slemrmartin: Theoretically it can be used in the future so we don't need to remove it now, thus we should keep this concern for this model

app/models/concerns/availability_status_concern.rb
pkomanek: That’s a good idea. Requested code has been imported. Thanks

lindgrenj6 (approved):

PR #296, opened 2021-01-06
Pass on headers that come through the Availability Status Listener

lindgrenj6:
The availability status listener wasn't passing on headers, this fixes that. 

\# DEPENDS ON:
- [x] https://github.com/RedHatInsights/topological_inventory-providers-common/pull/73
- [x] common gem release

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -51,7 +51,7 @@ class AvailabilityStatusListener
     options[:last_available_at] = options[:last_checked_at] if options[:availability_status] == 'available'
 
     object.update!(options)
-    raise_event("#{model_class}.update", object.as_json)
+    raise_event("#{model_class}.update", object.as_json, event.headers)
   rescue NameError
     Rails.logger.error("Invalid resource_type #{payload["resource_type"]}")
   rescue ActiveRecord::RecordNotFound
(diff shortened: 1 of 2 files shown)


lib/availability_status_listener.rb
```diff
@@ -62,8 +62,8 @@ def process_event(event)
     Rails.logger.error(["Something is wrong when processing Kafka message: ", e.message, *e.backtrace].join($RS))
   end
 
-  def raise_event(event, payload)
-    Sources::Api::Events.raise_event(event, payload)
+  def raise_event(event, payload, headers)
```
syncrou, line 65: Is there a chance this should be `headers=nil` by default?

lindgrenj6: I debated that, but the sources event class already handles it for us:
https://github.com/lindgrenj6/sources-api/blob/8e904b4371e2d716061155cc30978385feae3a2c/lib/sources/api/events.rb#L4

syncrou: @lindgrenj6 makes sense.  Sounds good.

syncrou (approved):
LGTM 👍

slemrmartin (approved):
:+1:

lib/availability_status_listener.rb
```diff
@@ -51,7 +51,7 @@ def process_event(event)
     options[:last_available_at] = options[:last_checked_at] if options[:availability_status] == 'available'
 
     object.update!(options)
-    raise_event("#{model_class}.update", object.as_json)
+    raise_event("#{model_class}.update", object.as_json, event.headers)
```
slemrmartin, line 54: @lindgrenj6 does `event.headers` return nil or raise exception if not set in operations worker?

lindgrenj6: No, it is just an empty hash luckily.

PR #297, opened 2021-01-08
Switch to using common gem's `params_for_update`

lindgrenj6:
After adding `:extra` to the `Application` model it was reported by @dccurtis that patch-ing extra wasn't working.

Digging into it a bit resulted in having to switch to using the common gem's `params_for_update` method from the Parameters mixin. 

---

This change also exposed some problems in the common gem method as well, so there is a subsequent PR over there that will be linked here.

\# DEPENDS ON:
- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/213
- [x] https://github.com/RedHatInsights/insights-api-common-rails/pull/215

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
 gem 'cloudwatchlogger',     '~> 0.2.1'
-gem 'insights-api-common',  '~> 4.0'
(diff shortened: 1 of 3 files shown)


slemrmartin (commented):
@lindgrenj6 can you give me an example what this PR solves?

lindgrenj6:
@slemrmartin as we talked in DM, basically switching over sources-api to use the common gem's handling for `params_for_update`. Going to update it in Topo API as well so we aren't changing behavior in various controllers. 

image of fix: 
image

slemrmartin (approved):

PR #299, opened 2021-01-15
Google Seeds

slemrmartin:
Source Type and ApplicationType seeds for Google Cloud Engine

* [x] **depends on** https://github.com/RedHatInsights/platform-mq/pull/114

---

[RHCLOUD-11249](https://issues.redhat.com/browse/RHCLOUD-11249)

### DIFF
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -124,25 +124,32 @@ ansible-tower:
           :when: endpoint.verify_ssl
           :is: true
 google:
-  :product_name: Google Cloud Engine
+  :product_name: Google Cloud
   :vendor: Google
-  :icon_url: "/apps/chrome/assets/images/partners-icons/google-cloud-engine.svg" # TODO
+  :icon_url: "/apps/frontend-assets/partners-icons/google-cloud.svg"
   :schema:
     :authentication:
-    - :type: username_password
+    - :type: project_id_service_account_json
       :name: Project ID and Service Account JSON
       :fields:
       - :component: text-field
         :name: authentication.authtype
         :hideField: true
         :initializeOnMount: true
-        :initialValue: username_password
+        :initialValue: project_id_service_account_json
       - :component: text-field
         :name: authentication.username
         :label: Project ID
       - :component: textarea
         :name: authentication.password
         :label: Service Account JSON
+      - :component: text-field
+        :name: application.extra.dataset
+        :stepKey: cost-management
+        :label: Dataset name
+        :isRequired: true
+        :validate:
+        - :type: required
     :endpoint:
       :hidden: true
       :fields:
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -123,6 +123,34 @@ ansible-tower:
         :condition:
           :when: endpoint.verify_ssl
           :is: true
+google:
+  :product_name: Google Cloud Engine
+  :vendor: Google
+  :icon_url: "/apps/chrome/assets/images/partners-icons/google-cloud-engine.svg" # TODO
+  :schema:
+    :authentication:
+    - :type: username_password
+      :name: Project ID and Service Account JSON
+      :fields:
+      - :component: text-field
+        :name: authentication.authtype
+        :hideField: true
+        :initializeOnMount: true
+        :initialValue: username_password
+      - :component: text-field
+        :name: authentication.username
+        :label: Project ID
+      - :component: textarea
+        :name: authentication.password
+        :label: Service Account JSON
+    :endpoint:
+      :hidden: true
+      :fields:
+      - :component: text-field
+        :name: endpoint.role
+        :hideField: true
+        :initializeOnMount: true
+        :initialValue: google
 azure:
   :product_name: Microsoft Azure
   :vendor: Azure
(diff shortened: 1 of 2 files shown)


slemrmartin:
cc @dccurtis

slemrmartin:
@dccurtis availability check url will be the same like for aws? `/api/cost-management/v1/source-status/`

dccurtis:
> @dccurtis availability check url will be the same like for aws? `/api/cost-management/v1/source-status/`

Yes, no difference there

db/seeds/source_types.yml
```diff
@@ -123,6 +123,32 @@ ansible-tower:
...
           :when: endpoint.verify_ssl
           :is: true
+google:
+  :product_name: Google Cloud Engine
+  :vendor: Google
+  :icon_url: "/apps/chrome/assets/images/partners-icons/google-cloud-engine.svg" # TODO
+  :schema:
+    :authentication:
+    - :type: username_password
+      :name: Project ID and Service Account JSON
```
dccurtis, line 133: is the Service Account JSON going to be used to seed the service account email for the wizard?

db/seeds/source_types.yml
```diff
@@ -123,6 +123,32 @@ ansible-tower:
...
+      :fields:
+      - :component: text-field
+        :name: authentication.authtype
+        :hideField: true
+        :initializeOnMount: true
+        :initialValue: username_password
```
dccurtis, line 139: Just want to make sure I understand: `project_id` will go in the `Authentications` `authtype` field?

db/seeds/source_types.yml
```diff
@@ -123,6 +123,32 @@ ansible-tower:
...
+        :label: Project ID
+      - :component: textarea
+        :name: authentication.password
+        :label: Service Account JSON
+    :endpoint:
```
dccurtis, line 144: Would we be able to leverage the new `Application` `extra` field to store the `dataset`?

Something like this
```
      - :name: application.extra.dataset
        :stepKey: cost-management
        :component: text-field
        :label: Dataset name
        :isRequired: true
```

db/seeds/source_types.yml
slemrmartin: no, that's mistake :)

db/seeds/source_types.yml
slemrmartin: I don't know, but in the credentials provided by QE there is a `client_email` key-value in the JSON

slemrmartin: Anyway, I started from the ManageIQ's create form: https://files.gitter.im/5aa15e4ad73408ce4f907617/dsFs/Screenshot-2021-01-15-at-10.44.16.png

dccurtis: Are there other use cases for GCP that you are working on besides Cost Management?

slemrmartin: There is and oportunity for topological-inventory collector, but we don't have any milestone for that as far as i know

db/seeds/source_types.yml
rvsia: Yes, I agree! 👍  We will do it anyway in the future and doing it now will prevent any painful migration.

(You should also provide a validate in that object:

```
        :validate:
        - :type: required
```

`isRequired` is just a boolean to show the required marker.)

db/seeds/source_types.yml
```diff
@@ -123,6 +123,34 @@ ansible-tower:
         :condition:
           :when: endpoint.verify_ssl
           :is: true
+google:
+  :product_name: Google Cloud Engine
+  :vendor: Google
+  :icon_url: "/apps/chrome/assets/images/partners-icons/google-cloud-engine.svg" # TODO
```
rvsia, line 129: I have already uploaded the logo: '/apps/frontend-assets/partners-icons/google-cloud.svg'

db/seeds/source_types.yml
```diff
@@ -123,6 +123,34 @@ ansible-tower:
         :condition:
           :when: endpoint.verify_ssl
           :is: true
+google:
+  :product_name: Google Cloud Engine
```
rvsia, line 127: I think we should discuss with UX how to call it, this value is then shown in the ui in multiple places. @beccaglass

db/seeds/source_types.yml
```diff
@@ -123,6 +123,34 @@ ansible-tower:
...
+  :vendor: Google
+  :icon_url: "/apps/chrome/assets/images/partners-icons/google-cloud-engine.svg" # TODO
+  :schema:
+    :authentication:
+    - :type: username_password
```
rvsia, line 132: I think we should describe it more accurately as we don't actually provide any username or password - `project_id` maybe ?.

db/seeds/source_types.yml
beccaglass: My preference is 'Google Cloud' -- seems to be the most scalable

db/seeds/source_types.yml
slemrmartin: added

db/seeds/source_types.yml
slemrmartin: changed to `project_id_service_account_json`

rvsia (approved):
:+1:

dccurtis:
Thanks for adding the extra field for dataset!  I'll update on my side to query for `project_id_service_account_json` to get the `project_id` from the username field 👍

lindgrenj6 (approved):

PR #301, opened 2021-01-19
Update Ruby stream from 2.5 -> 2.6

lindgrenj6:
Just bumping the ruby module version from 2.5 -> 2.6, been meaning to do this for a while but a bug in the ruby2_keywords gem is leading to this rather than pinning them gem. We have been testing against 2.6 in travis for months now so it should be pretty safe. (I'm opening this on all repos, if we don't want to do this I am completely fine with closing this for now.)


syncrou (approved):
LGTM 👍

PR #302, opened 2021-01-20
Add meta_data table + models + seeding

lindgrenj6:
Heres the initial modeling for AppMetaData ~`AppMetaDatum` (latin anyone? since data is apparently plural)~. 

EDIT: ended up making MetaData a STI class. We now have 2 subclasses: `SuperKeyMetaData` and `AppMetaData`. They will both be seeded. 

- I set up a manual `seed` method on both since the seeding concern won't work for this type of stuff
- turns out you can embed `json` directly into `yaml`, so we can directly support the JSON payloads in the superkey seeding file. 
- there is an association on `ApplicationType` for both `*meta_data`'s for easy access
- the `name` enum is what defines what we do/do not support for SuperKeyMetaData, can expand as necessary (or the more we support)
- the `substitutions` column is what holds the key-value pairs of what should be substituted out of the JSON string payload. 

e2e-deploy PR here for `SOURCES_ENV` required for AppMetaData seeding: https://github.com/RedHatInsights/e2e-deploy/pull/2681

### DIFF
--- a/db/seeds/app_metadata.yml
+++ b/db/seeds/app_metadata.yml
@@ -1,0 +1,19 @@
+######
+# Seed file for application specific metadata.
+# Outer object is the env (in case there are per-env settings)
+# Innter object is in the form of `key`: `jsonb`, but a plain
+# string is valid json. So plain key/value works fine.
+#####
+
+ci:
+  "/insights/platform/cost-management":
+    gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
+qa:
+  "/insights/platform/cost-management":
+    gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
+stage:
+  "/insights/platform/cost-management":
+    gcp_service_account: ccc
+prod:
+  "/insights/platform/cost-management":
+    gcp_service_account: ddd
--- a/app/models/app_meta_data.rb
+++ b/app/models/app_meta_data.rb
@@ -1,0 +1,24 @@
+class AppMetaData < MetaData
+  belongs_to :application_type
+
+  def self.seed
+    transaction do
+      env = ENV['SOURCES_ENV']
+      raise "No SOURCES_ENV set, cannot continue seeding." unless env
+
+      destroy_all
+
+      metadata = YAML.safe_load(File.read("db/seeds/app_metadata.yml"))
+      metadata[env].each do |app, settings|
+        apptype = ApplicationType.find_by(:name => app)
+
+        settings.each do |key, value|
+          apptype.app_meta_data.create!(
+            :name    => key,
+            :payload => value
+          )
+        end
+      end
+    end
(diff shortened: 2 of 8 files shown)


app/models/app_meta_datum.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def self.seed
+    transaction do
+      YAML.safe_load(File.read("db/seeds/superkey_metadata.yml"))
+          .deep_symbolize_keys
+          .each do |app, settings|
+        apptype = ApplicationType.find_by(:name => app)
+
+        settings[:steps].each do |step|
+          AppMetaDatum.create!(
```
syncrou, line 14: I don't mind the naming.  I'm okay with it if we don't expose it via the API - I'm concerned if we do that the naming will be slightly confusing ( even if the use of `Datum` is correct. ) But I'm not going to split hairs over it.

syncrou (commented):
With my quick overview I like it. Not so sure on the naming, though I get the reasons for `Datum` with Rails default to singular model names.  Outside of that it looks great.

app/models/app_meta_datum.rb
slemrmartin: I'm for rename, it's confusing to me

lindgrenj6:
This is ready. Decided against including the endpoint stuff in this PR, going to branch off of this to add the index/post/patch endpoints.

app/models/app_meta_datum.rb
lindgrenj6: So I'm all for changing the name as well, but running into an issue with the common gem here: https://github.com/RedHatInsights/insights-api-common-rails/blob/12078a13696efe07beddc3a97f173f7cdde47e5f/lib/insights/api/common/application_controller_mixins/api_doc.rb#L32

Basically we are calling `singularize` on everything which makes sense. Would we be ok with my overriding this method in the sources ApplicationController? Since this is a one-off. It's either that or use the funky name in the openapi but _not_ in the application.

db/seeds/app_metadata.yml
```diff
@@ -0,0 +1,19 @@
...
+# Innter object is in the form of `key`: `jsonb`, but a plain
+# string is valid json. So plain key/value works fine.
+#####
+
+ci:
+  "/insights/platform/cost-management":
+    gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
+qa:
+  "/insights/platform/cost-management":
+    gcp_service_account: bbb
```
dccurtis, line 13: Can use `test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com` here too 👍

syncrou (commented):
This looks good to me - Per @dccurtis comment - it looks like we can use a real billing account instead of the placeholder.

lindgrenj6:
@dccurtis @syncrou added gcp setting to qa.

PR #306, opened 2021-01-22
Add app_meta_data endpoints

lindgrenj6:
PR for adding all the endpoints for `ApplicationType` `MetaData` models.

After thinking it through, we don't want these to be editable in any way (except maybe /internal, but they're gonna be public anyway), so I'm going to add some seeding for the app meta data type. 

\# TODO:
- [x] index
~- [x] post~
~- [x] patch~
~- [x] delete~
- [x] openapi working with all
- [x] tests

### DIFF
--- a/spec/requests/api/v3.1/app_meta_data_spec.rb
+++ b/spec/requests/api/v3.1/app_meta_data_spec.rb
@@ -1,0 +1,37 @@
+RSpec.describe("v3.1 - AppMetaData") do
+  include ::Spec::Support::TenantIdentity
+
+  let(:messaging_client) { instance_double("ManageIQ::Messaging::Client") }
+  before do
+    allow(messaging_client).to receive(:publish_topic)
+    allow(Sources::Api::Messaging).to receive(:client).and_return(messaging_client)
+  end
+
+  let(:headers)          { {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
+  let(:attributes)       { {"name" => "a_field", "payload" => {"value" => 1234}} }
+  let(:collection_path)  { "/api/v3.1/app_meta_data" }
+
+  describe("/api/v3.1/app_meta_data") do
+    context "get" do
+      it "success: empty collection" do
+        get(collection_path, :headers => headers)
+
+        expect(response).to have_attributes(
+          :status      => 200,
+          :parsed_body => paginated_response(0, [])
+        )
+      end
+
+      it "success: non-empty collection" do
+        create(:app_meta_data, attributes)
+
+        get(collection_path, :headers => headers)
+
+        expect(response).to have_attributes(
+          :status      => 200,
+          :parsed_body => paginated_response(1, [a_hash_including(attributes)])
+        )
+      end
+    end
+  end
+end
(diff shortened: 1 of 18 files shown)


slemrmartin (commented):
I think we can move it to the v3.x together with #308

app/controllers/api/v1x0.rb
```diff
@@ -13,5 +13,6 @@ class EndpointsController < Api::V1::EndpointsController; end
     class GraphqlController < Api::V1::GraphqlController; end
     class SourcesController < Api::V1::SourcesController; end
     class SourceTypesController < Api::V1::SourceTypesController; end
+    class AppMetaDataController < Api::V1::AppMetaDataController; end
```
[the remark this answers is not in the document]
lindgrenj6: No, we do not. Initially I was doing this in case we wanted to have the field GET-able on all versions, but we have decided against it. Working on removing it now.

spec/requests/api/v3.1/app_meta_data_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  include ::Spec::Support::TenantIdentity
+
+  let(:messaging_client) { instance_double("ManageIQ::Messaging::Client") }
+  before do
+    allow(messaging_client).to receive(:publish_topic)
+    allow(Sources::Api::Messaging).to receive(:client).and_return(messaging_client)
+  end
+
+  let(:headers)          { {"CONTENT_TYPE" => "application/json", "x-rh-identity" => identity} }
+  let(:attributes)       { {"name" => "a_field", "payload" => {"value" => 1234}} }
```
syncrou, line 11: This question may be off base - but if we have an STI class AppMetaData - wouldn't the payload include a "type" key? - Maybe it's supressed in openapi spec and this question is mute.  

I'm coming from the perspective of a DB lookup:  `AppMetaData.last` would return the `:type` attribute `==` `app_meta_data`. So I'm wondering if the returned json payload would include a "type" key?

syncrou:
@roliveri can you take a look as well?

spec/requests/api/v3.1/app_meta_data_spec.rb
lindgrenj6: Yep you're right, it won't unless we expose it in the openapi.json. We can expose it if we want to, but this is the only "user facing" metadata so a type doesn't quite make sense unless we expand it. 

Current output here:
```
http :3002/api/sources/v3.1/app_meta_data $rhid --body
{
    "data": [
        {
            "application_type_id": "2",
            "created_at": "2021-01-26T21:08:09Z",
            "id": "81",
            "name": "gcp_service_account",
            "payload": "test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com",
            "updated_at": "2021-01-26T21:08:09Z"
        }
    ],
    "links": {
        "first": "/api/sources/v3.1/app_meta_data?offset=0",
        "last": "/api/sources/v3.1/app_meta_data?offset=0"
    },
    "meta": {
        "count": 1,
        "limit": 100,
        "offset": 0
    }
}
```

syncrou: No, I'm fine with it not being exposed - unless the SuperKeyMetaData would be exposed: In that case the type would be helpful and/or would most likely be a required field for a POST.

For now this makes sense to me.

syncrou (approved):
LGTM 👍

roliveri (approved):

PR #308, opened 2021-01-25
Add 'app_creation_workflow' enumerated string to Source to track creation type

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-11657

Adding this field to track what type of workflow we should use when creating applications off of said source. It's an enum (in the openapi and in the rails model validation) of `%w(trust semi-trust paranoid)`.

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -11,7 +11,7 @@ class Source < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available partially_available unavailable] }, :allow_nil => true
-  validates :app_creation_workflow, :inclusion => {:in => %w[paranoid semi-trust trust]}
+  validates :app_creation_workflow, :inclusion => {:in => %w[manual_configuration account_authorization]}
 
   belongs_to :source_type
 
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -11,6 +11,7 @@ class Source < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available partially_available unavailable] }, :allow_nil => true
+  validates :app_creation_workflow, :inclusion => {:in => %w[paranoid semi-trust trust]}
 
   belongs_to :source_type
 
--- a/db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
+++ b/db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
@@ -1,5 +1,5 @@
 class AddAppCreationWorkflowToSources < ActiveRecord::Migration[5.2]
   def change
-    add_column :sources, :app_creation_workflow, :string, :default => "paranoid"
+    add_column :sources, :app_creation_workflow, :string, :default => "manual_configuration"
   end
 end
--- a/db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
+++ b/db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
@@ -1,0 +1,5 @@
+class AddAppCreationWorkflowToSources < ActiveRecord::Migration[5.2]
+  def change
+    add_column :sources, :app_creation_workflow, :string, :default => "paranoid"
+  end
+end
--- a/public/doc/openapi-3-v3.1.json
+++ b/public/doc/openapi-3-v3.1.json
@@ -2098,6 +2098,14 @@
       "Source": {
         "type": "object",
         "properties": {
+          "app_creation_workflow": {
+            "type": "string",
+            "enum": [
+              "paranoid",
+              "semi-trust",
+              "trust"
+            ]
+          },
           "availability_status": {
             "type": "string"
           },
--- a/public/doc/openapi-3-v3.1.json
+++ b/public/doc/openapi-3-v3.1.json
@@ -2101,9 +2101,8 @@
           "app_creation_workflow": {
             "type": "string",
             "enum": [
-              "paranoid",
-              "semi-trust",
-              "trust"
+              "manual_configuration",
+              "account_authorization"
             ]
           },
           "availability_status": {


lindgrenj6:
One question: should we have this attribute in _all_ api versions? I was thinking only including most of these superkey fields in the latest.

lindgrenj6:
cc @syncrou

syncrou (commented):
This looks good so far.

syncrou:
@roliveri - Do you have any thoughts on api versioning for this change? https://github.com/RedHatInsights/sources-api/pull/308#issuecomment-766996259 - Since it's a model update there is no way to turn off the validation regardless of the api version - but the open api spec at the very least would not expose the attribute as shown in this PR.  

@lindgrenj6 potentially specs would be able to prove what would happen with openapi v1 vs v3.x?

public/doc/openapi-3-v3.0.json
```diff
@@ -1914,6 +1914,15 @@
       "Source": {
         "type": "object",
         "properties": {
+          "app_creation_workflow": {
```
syncrou, line 1917: Should this be a 3.x change instead of added to v3?

roliveri:
So, for older versions of the API, would this attribute have to default to something reasonable on create?

For the same older versions, should the attribute be returned on get?

lindgrenj6:
@syncrou @roliveri  if we left the field out of all the other openapi schemas (1.0 - 3.0) the field would just default to `paranoid` on create, and it would not be returned in the GET responses for those older versions. 

I think if we are going to go the new version route (like v3.1), we'll have all the new superkey-fields and operations in that version but not in the older ones. The UI (and customers using it) would just always have to use the latest version to get the super-key related fields or to use any of these newer endpoints.

public/doc/openapi-3-v3.0.json
lindgrenj6: Yes I like this idea. I will change to v3.1 in my other index PR and update this one with those changes.

public/doc/openapi-3-v3.0.json
```diff
@@ -1914,6 +1914,15 @@
...
+            "type": "string",
+            "readOnly": true,
```
slemrmartin, line 1919: should it be read only? customer should be able to change his/her mind and switch it, shouldn't (s)he?

db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
```diff
@@ -0,0 +1,5 @@
+class AddAppCreationWorkflowToSources < ActiveRecord::Migration[5.2]
+  def change
+    add_column :sources, :app_creation_workflow, :string, :default => "paranoid"
```
slemrmartin, line 3: what about to switch to postgres enum?
i.e. https://medium.com/@diegocasmo/using-postgres-enum-type-in-rails-799db99117ff

public/doc/openapi-3-v3.0.json
lindgrenj6: Ahhh yeah thats a good point. I was just thinking on initial source creation they would pick the workflow and their applications and be done. They could come back and be like "I don't really feel like doing the wizard again, do it for me".

I'll get that changed.

db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
lindgrenj6: Initially I thought about doing that, but we did it for something similar on Catalog and it ended up causing nothing but headaches since AR doesn't play very nicely with the postgres enum type. Things get tough any time you need to change the data or add to it and it makes some AR operations not quite work (we had problems with rake db:rollback etc if I remember right) 

At the end of the day this is a pretty low-traffic table, the field only gets validated on create/update so I think keeping it a simple string would be better, the validation will make it behave as an enum anyway.

public/doc/openapi-3-v3.0.json
lindgrenj6: I'll rebase this PR onto master once #306 is merged.

roliveri (approved):

app/models/source.rb
```diff
@@ -11,6 +11,7 @@ class Source < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available partially_available unavailable] }, :allow_nil => true
+  validates :app_creation_workflow, :inclusion => {:in => %w[paranoid semi-trust trust]}
```
rvsia, line 14: I would change these values to something more neutral, as it is can be still read by users in their network tabs.

`manual semi-auto auto`

or we can call it the same as in the UI

`manual-configuration account-authorization` 

Also, there is no semi-trust mode, maybe we could skip for now.

slemrmartin: I would like to have these values equal with UI to avoid double-naming confusion, sync with UX could be helpful

lindgrenj6: Will update it to match the UI, we can add semi-trust when necessary.

app/models/source.rb
```diff
@@ -11,6 +11,7 @@ class Source < ApplicationRecord
 
   attribute :availability_status, :string
   validates :availability_status, :inclusion => { :in => %w[available partially_available unavailable] }, :allow_nil => true
+  validates :app_creation_workflow, :inclusion => {:in => %w[manual-configuration account-authorization]}
```
rvsia, line 14: Oh, sorry I wrote it wrong, but we should keep it consistent with availibility_status and use `_` too (`manual_configuration` etc)

db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
rvsia, line 3: The default should be changed to the new string.

rvsia (approved):

db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb
syncrou: @slemrmartin enum makes a lot of sense, except that AR and postgres enum together are a disaster. I implemented enum in Catalog ( like @lindgrenj6 references ) and we fought it the entire time after.  

I am for keeping this simpler string based validation as AR will do the right thing.

slemrmartin (approved):

rvsia:
@lindgrenj6 I think this issue https://issues.redhat.com/browse/RHCLOUD-11703 is a duplicate.

PR #310, opened 2021-01-27
Add Superkey field to ApplicationType and SourceType

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-11704

So this is a fun one. We have to find a way to mark certain Applications as supporting superkey (like we will probably start with _just_ subs-watch superkey through amazon, not for cost. That will come later).

So the thought is marking the SourceType Authentication for aws `access_key_secret_key` as `superkey_supported`, so for AWS sources we know to use that auth type. 
We also need to mark the cost/subswatch amazon auth types as superkey supported, so we look that amazon supports superkey, then look back at the source type to figure out which auth type to use.

Definitely open to discussion on this one. @rvsia and I went back and forth this morning about it a bit already.


db/seeds/application_types.yml
```diff
@@ -19,6 +19,7 @@
   :supported_authentication_types:
     amazon:
     - arn
+    - superkey
```
rvsia, line 22: So as I proposed in DMs, we don’t need to store which applications support the superkey, as we should support the feature for whole types. (Now, as we are talking just about AWS type, both its applications shares the same config credentials, so there will be not so much work more to do to support both of them.)

db/seeds/application_types.yml
```diff
@@ -33,6 +34,7 @@
   :supported_authentication_types:
     amazon:
     - cloud-meter-arn
+    - superkey
```
rvsia, line 37: ^

db/seeds/source_types.yml
```diff
@@ -4,6 +4,7 @@ amazon:
   :schema:
     :authentication:
     - :type: access_key_secret_key
+      :superkey_suported: true
```
rvsia, line 7: What about `:is_superkey`? Superkey_supported sounds more like an attribute of a source type.

db/seeds/application_types.yml
lindgrenj6: Yep, sorry for taking so long on this, it also doesn't make sense to have this here because the superkey authentication is based on the `Source` record, not the application. Whoops!

db/seeds/source_types.yml
lindgrenj6: Sure, I'm ok with it since you'll be keying off of it more than me haha.

db/seeds/source_types.yml
```diff
@@ -4,6 +4,7 @@ amazon:
   :schema:
     :authentication:
     - :type: access_key_secret_key
+      :is_superkey?: true
```
rvsia, line 7: Oh, the `?` would not be much friendly for javascript code. :sweat_smile:  I know it's a Ruby standard to use it for booleans, but maybe we can make an exception here. :smile:

syncrou: @lindgrenj6 thoughts on this?

syncrou (commented):
Looks good, except @rvsia has a good question that should probably be resolved prior to merging.

db/seeds/source_types.yml
lindgrenj6: Yep updated, my bad I immediately put `?` at the end of booleans!

roliveri (approved):

app/models/app_meta_data.rb
```diff
@@ -3,8 +3,8 @@ class AppMetaData < MetaData
 
   def self.seed
     transaction do
-      env = ENV['SOURCES_ENV']
-      raise "No SOURCES_ENV set, cannot continue seeding." unless env
+      # Default to CI
+      env = ENV['SOURCES_ENV'] || "ci"
```
slemrmartin, line 7: there will be mandatory ENV_NAME parameter in ClowdApp CRD so I'm wondering if we can join these env values.
But it would need to have some default because the ENV_NAME can be "dev-env" , "ephemeral"(?) etc.
So I want to discuss about it

rvsia (approved):

app/models/app_meta_data.rb
lindgrenj6: maybe we can just make it default to `ci` if it isn't found. I like that `ENV_NAME` is going to be present - we can just remove `SOURCES_ENV` from the e2e-deploy deploymentconfig that way.

slemrmartin (approved):

PR #313, opened 2021-02-02
updates to work with superkey worker

lindgrenj6:
This PR contains the schema changes I did when creating the initial superkey worker: https://github.com/lindgrenj6/sources-superkey-worker/

Basically I just added a few fields to the app meta data, and changed the schema a bit on the superkey metadata. I also added a helper method for producing a request to the superkey worker. 

I am continuing the work in #314, which uses said method and app meta data fields.

### DIFF
--- a/app/models/authentication.rb
+++ b/app/models/authentication.rb
@@ -19,6 +19,10 @@ class Authentication < ApplicationRecord
   private
 
   def set_source
-    self.source_id = resource&.source_id
+    self.source_id = if resource.instance_of?(Source)
+                       resource.id
+                     else
+                       resource.try(:source_id)
+                     end
   end
 end
--- a/db/seeds/app_metadata.yml
+++ b/db/seeds/app_metadata.yml
@@ -8,6 +8,9 @@
 ci:
   "/insights/platform/cost-management":
     gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
+    aws_wizard_account_number: "8675309"
+  "/insights/platform/cloud-meter":
+    aws_wizard_account_number: "8675309"
 qa:
   "/insights/platform/cost-management":
     gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
--- a/db/seeds/superkey_metadata.yml
+++ b/db/seeds/superkey_metadata.yml
@@ -3,10 +3,8 @@
   steps:
   - step: 1
     name: s3
-    payload:
-      bucket_name: S3
-    substitutions:
-      S3: gen_s3_name
+    payload: {}
+    substitutions: {}
   - step: 2
     name: policy
     payload:
@@ -37,7 +35,7 @@
         ]
       }
     substitutions:
-    - S3BUCKET: s3
+      S3BUCKET: s3
   - step: 3
     name: role
     payload:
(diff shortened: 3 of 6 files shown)


db/seeds/app_metadata.yml
```diff
@@ -8,6 +8,9 @@
 ci:
   "/insights/platform/cost-management":
     gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
+    aws_wizard_account_number: "8675309"
```
syncrou, line 11: I'm guessing this is fine with Cost to make static?  Should it be an env variable?

db/seeds/app_metadata.yml
```diff
@@ -8,6 +8,9 @@
...
+  "/insights/platform/cloud-meter":
+    aws_wizard_account_number: "8675309"
```
syncrou, line 13: Same.

db/seeds/superkey_metadata.yml
```diff
@@ -55,11 +53,11 @@
         ]
       }
     substitutions:
-    - ACCOUNT: get_account
+      ACCOUNT: get_account
   - step: 3
     name: bind_role
     payload: {}
-    substitutions: []
+    substitutions: {}
```
syncrou, line 60: Just for my curiosity, why did substitutions change from an `[]` to a `{}`?

syncrou (commented):
This looks great. Some small questions on my part.  @slemrmartin would be great to get your input.

db/seeds/app_metadata.yml
lindgrenj6: Well, I will substitute it out eventually for the real account, but that is a dummy number for now (jenny jenny...)

The number is public and in the docs + wizard, so it will probably be ok to be public.

db/seeds/superkey_metadata.yml
lindgrenj6: Basically I [changed my schema](https://github.com/lindgrenj6/sources-superkey-worker/blob/03fc9fd558b1a23f4917d17b42179f40ce8ffea5/provider/types.go#L22) on the superkey side to be a `map[string]string` instead of an `[]map[string]string`, since each item was a single k/v pair to begin with. It makes more sense to iterate over k/v in the map than to iterate over an array of single k/v pairs.

db/seeds/app_metadata.yml
roliveri: I feel sorry for the people who actually had that phone number.

Does this need to be changed before it's merged, or does it not matter?

lindgrenj6: Nope, it will be fine for now. I will update it when the superkey worker is deployed and ready, that way we aren't accidentally creating resources in the meantime.

roliveri (approved):

app/models/authentication.rb
```diff
@@ -19,6 +19,10 @@ class Authentication < ApplicationRecord
   private
 
   def set_source
-    self.source_id = resource&.source_id
+    self.source_id = if resource.instance_of?(Source)
+                       resource.id
```
slemrmartin, line 23: @lindgrenj6 when this happens?

slemrmartin: btw using `resource_type` and `resource_id` saves one DB request

lindgrenj6: This happens when using the `Source` as the `resource` for an Authentication. Which I hope is what we're going to be doing for superkey authentications that way we don't have to have a dummy endpoint or anything like that involved.

PR #314, opened 2021-02-03
add BulkCreate controller and openapi objects

lindgrenj6:
kind of relates to #313 but this is more on the API side.

\# TODO:
- [x] openapi spec for bulkcreate/bulkresponse
- [x] bulk create post action that creates ALL THE RESOURCES!
- [x] tests, lots and lots of tests

**UPDATE 2/9/21**
So we have the create workflow mostly done here. The bulk_create endpoints as well as the superkey post work with the worker, I just have to implement the teardown functionality which is already implemented but I have to add structs + deserialization for it. 

**UPDATE 2/11/21**
This is in a good spot. I have moved the superkey workflow to `after_create`/`after_destroy` callbacks, so the workflows just get called on normal resource operations.

I have to implement credential checking still, but I can do that in a followup PR.


syncrou (commented):
Is the Hakiri issue solvable?

mkanoor:
@lindgrenj6 Instead of application_id and source_type_id which are numeric values, can we pass them as strings like

 **Application**
  * /insights/platform/cost-management
  * /insights/platform/topological-inventory
  * /insights/platform/cloud-meter
  * /insights/platform/fifi
  * /insights/platform/catalog

**Source Type**
   * satellite
   * openshift
   * azure
   * ansible-tower
   * amazon

lindgrenj6:
@mkanoor yes, that's how I set it up. 

The `type` field in the `BulkCreatePayload` object of each allows you to just specify a string rather than the ID, I look it up on the backend. 

Application matching just does a `String#match?` on the name, so those strings work. 
SourceType is an `enum` in the openapi matching those strings.

app/controllers/api/v3x1/bulk_create_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Api
+  module V3x1
+    class BulkCreateController < ApplicationController
+      def create
+        # Authorize creating a new source, before we go through the processing.
+        authorize(Source.new)
```
mkanoor, line 6: @lindgrenj6 When we have an intermediary (e.g Cloud Controller) creating the Source object should it be building a temporary XRH with the account_number in it when it calls the end point? Do you support Pre Shared Keys (PSK) that can  be used to trust connections from services running inside the cloud.redhat.com.

lindgrenj6: We could support a PSK - but the x-rh-id is more for the account number than anything. 

If we wanted to do an internal bulk-create operation I could see us also offering a kafka consumer for that operation. Not sure if that's in the cards right now though, but we could do it eventually.

rvsia:
So as we discussed in our video call, we should use the `app_creation_workflow` (https://github.com/RedHatInsights/sources-api/pull/308) as a determinant if the worker should create the credentials for applications or skip it and does nothing (as the credentials are created by users).

The same should work when a new application is added: if a new application is added and the source is set to 'account_authorization' then the worker should start creating credentials, if not, it should do nothing.

You can use the new superKey request as an internal one, so it should not be so much big change for you and this change will greatly improve the experience. (The only difference between superkey and non-super key sources is that one `app_creation_workflow` attribute, the rest can stay the same.)

mkanoor:
@lindgrenj6 In the Source Object there is an attribute called Source_ref that is the unique identifier for the source like a Tower UUID. Currently when we check for the existence of the tower or satellite in Source we have to compare the Source_ref to check if the Source already exists. In the bulk create will there be any provision to add this check, or should the playbook installers for tower and satellite keep checking if the Source with Source_ref exists in the DB before creating a new one.

lindgrenj6:
@mkanoor 
>Currently when we check for the existence of the tower or satellite in Source we have to compare the Source_ref to check if the Source already exists.

Sounds like a validation to me and there currently is not a validation on that column. Maybe we can make it unique per tenant much like the name? Thoughts @syncrou @roliveri?

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        raise ActiveRecord::ActiveRecordError, "Source Type not found" if srct.nil?
+
+        Source.create!(source.merge!(:source_type => srct))
+      end
+    end
+
+    def create_endpoints(endpoints, resources)
+      endpoints&.map do |endpoint|
+        src = find_resource(resources, :sources, endpoint.delete(:source_name))
```
slemrmartin, line 58: what is the structure of payload? This column doesn't exist in the model so calling this operation needs undocumented knowledge

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        Endpoint.create!(endpoint.merge!(:source_id => src.id))
+      end
+    end
+
+    def create_applications(applications, resources)
+      applications&.map do |app|
+        src = find_resource(resources, :sources, app.delete(:source_name))
+        # Get the application by id or lookup by type string
+        appt = ApplicationType.find_by(:id => app.delete(:application_type_id)) || get_application_type(app.delete(:type))
```
slemrmartin, line 68: 1) the same question like in create_sources
2) it checks against ApplicationType.name so if it's necessary, it should be consistent with the `create_sources` method and should be called `app.delete(:application_type_name)`

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        ::Application.create!(app.merge!(:source_id => src.id, :application_type_id => appt.id))
+      end
+    end
+
+    def create_authentications(authentications, resources)
+      authentications&.map do |auth|
+        ptype = auth.delete(:resource_type)
```
slemrmartin, line 76: Can we use `resource_type` and `resource_name`? 
Variables `ptype` and `pname` are meaningless

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+        pname = auth.delete(:resource_name)
+
+        # complicated logic here - since the source/endpoint/application can all
+        # be looked up differently
+        parent = case ptype
+                 when "source"
```
slemrmartin, line 82: The polymorphic "type" column contains first big letter - "Source"

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+          # create the application_authentication relation if the parent was an
+          # application.
+          ApplicationAuthentication.create!(:authentication => newauth, :application => parent) if ptype == "application"
+        end
+      end
+    end
+
+    private
+
+    def find_resource(resources, rtype, rname, field = :name)
```
slemrmartin, line 103: the same like above: `rtype` and `rname` are meaningless

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+    def create_authentications(authentications, resources)
+      authentications&.map do |auth|
+        ptype = auth.delete(:resource_type)
+        pname = auth.delete(:resource_name)
+
+        # complicated logic here - since the source/endpoint/application can all
+        # be looked up differently
+        parent = case ptype
+                 when "source"
+                   find_resource(resources, :sources, pname)
```
slemrmartin, line 83: maybe the 2nd param should match the `resource_type` format, so "Source" instead of :sources

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+    def find_resource(resources, rtype, rname, field = :name)
+      # use the safe operator in the case of creating a subresource on an
+      # existing source
+      parent = resources[rtype]&.detect { |resource| resource.send(field) == rname }
+
+      # if the parent is a source, it's possible that it was already created in
+      # the db so we need to try and look it up
+      if parent.nil?
+        case rtype
+        when :sources
```
slemrmartin, line 112: same as above, should match the resource_type values format, for better understandability

app/models/application.rb
```diff
@@ -10,14 +10,43 @@ class Application < ApplicationRecord
...
   def source_must_be_compatible
-  	return if application_type.supported_source_types.include?(source.source_type.name)
-  	errors.add(:source, "of type: #{source.source_type.name}, is not compatible with this application type")
+    return if application_type.supported_source_types.include?(source.source_type.name)
+
+    errors.add(:source, "of type: #{source.source_type.name}, is not compatible with this application type")
+  end
+
+  def superkey_workflow
+    return unless source.app_creation_workflow == "account_authorization"
```
slemrmartin, line 30: if the string will be used on more places, I vote for constants :)

lib/sources/super_key.rb
```diff
@@ -0,0 +1,42 @@
...
+      src = Source.find(@source_id)
+
+      extra = {}.tap do |e|
+        case @provider
+        when "amazon"
+          # account number to substitute in resources
+          e[:account] = @application.application_type
+                                    .app_meta_data
+                                    .detect { |field| field.name == "aws_wizard_account_number" }
+                                    .payload
```
slemrmartin, line 21: the `.payload` can be called on nil

slemrmartin (changes requested):

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+        self
+      rescue => e
+        Rails.logger.error("Error bulk processing from payload: Sources: #{@sources}, Endpoints: #{@endpoints}, Applications: #{@applications}, Authentications: #{@authentications}. Error: #{e}")
+
+        raise
+      end
+    end
+
+    def create_sources(sources)
+      sources&.map do |source|
```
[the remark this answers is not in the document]
lindgrenj6: well luckily the openapi validator will catch it for us. it wont' let anything through that doesn't match the `BulkCreatePayload` schema. https://github.com/lindgrenj6/sources-api/blob/932ba4c3f5f45e20a8265742bd04e5fca6893018/app/controllers/api/v1/sources_controller.rb#L57-L84

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+        # get the source type by ID or by type string
+        srct = SourceType.find_by(:id => source.delete(:source_type_id)) || SourceType.find_by(:name => source.delete(:type))
```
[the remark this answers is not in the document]
lindgrenj6: Yes, that is how we're doing the lookups for `SourceType` and `ApplicationType`. Instead of passing in the id (and requiring another GET to get the ID). 

EDIT: I ended up renaming it to `source_type_name`.

lib/sources/bulk_assembly.rb
lindgrenj6: The structure is in the openapi.json, here: https://github.com/lindgrenj6/sources-api/blob/932ba4c3f5f45e20a8265742bd04e5fca6893018/public/doc/openapi-3-v3.1.json#L2234-L2348

lib/sources/bulk_assembly.rb
lindgrenj6: Updating, for reference it was short for parent_type, parent_name, etc.

lib/sources/bulk_assembly.rb
lindgrenj6: Well this is coming through the payload, `resource_type` is where to attach the Authentication. I can capitalize it, but don't really want to require it to be capitalized in the payload.

lib/sources/bulk_assembly.rb
lindgrenj6: Updating, but for reference it was resource_type, resource_name.

lib/sources/bulk_assembly.rb
lindgrenj6: Sure, I usually prefer symbols to strings but it's the same either way.

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+          else
+            output[:authentications] = create_authentications(@authentications, output)
+          end
+        end
+
+        self
+      rescue => e
+        Rails.logger.error("Error bulk processing from payload: Sources: #{@sources}, Endpoints: #{@endpoints}, Applications: #{@applications}, Authentications: #{@authentications}. Error: #{e}")
+
+        raise
```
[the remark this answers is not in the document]
lindgrenj6: Since I'm raising ActiveRecordError, it gets wrapped as 400.

lib/sources/bulk_assembly.rb
lindgrenj6: Renamed to `application_type_name` for the lookup column.

app/models/application.rb
lindgrenj6: I think this might be the only place but I constantized it anyway.

lib/sources/bulk_assembly.rb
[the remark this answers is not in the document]
slemrmartin: I see, not sure about your link, you probably thought this one: https://github.com/RedHatInsights/sources-api/pull/314/files#diff-8e8060c901b3f6bc561ddb5cc297d413ae3b30eabb8bdef972d62582268335e7R2234

lib/sources/bulk_assembly.rb
```diff
@@ -0,0 +1,130 @@
...
+    def find_resource(resources, resource_type, resource_name, field = :name)
+      # use the safe operator in the case of creating a subresource on an
+      # existing source
+      parent = resources[resource_type]&.detect { |resource| resource.send(field) == resource_name }
+
+      # if the parent is a source, it's possible that it was already created in
+      # the db so we need to try and look it up
+      if parent.nil?
+        case resource_type
+        when "Sources"
```
slemrmartin, line 112: sorry I'm detailist, but I meant to be equal with the resource_type values which are singular

lib/sources/bulk_assembly.rb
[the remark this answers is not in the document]
lindgrenj6: Yep, clipboard error on my part. That second link is right.

syncrou (commented):
Looks great - 1 small nit.

syncrou (approved):
WIth all the applied feedback, going to accept this in its current form. LGTM 👍

syncrou:
Any issues missed in feedback can be requested as separate fixes and discussed from there.  Needing to finish the bulk create API side so UI can continue moving forward.

app/models/source.rb
```diff
@@ -24,4 +24,8 @@ def default_endpoint
     default = endpoints.detect(&:default)
     default || endpoints.build(:default => true, :tenant => tenant)
   end
+
+  def super_key
+    authentications.find_by!(:authtype => "superkey")
```
rvsia, line 29: Shouldn’t this be something like `find auth with a type thas has is_superkey set to true`?

lindgrenj6: I could see that being necessary if there was a use-case of having multiple superkeys on a source - but basically this is assuming only one SuperKey authentication will be created off of the source with `authtype` set to `superkey`, which should be the case. It will error out if there isn't a superkey authentication under the source. 

If we do want to allow multiple superkey sets of credentials per source we can revisit it, but I don't see a use-case for it. 

----

If we don't want to use `authtype => superkey` (which i didn't think of as an option until now) then I'd have to do a more sophisticated lookup, I can change it if we still want to use `authtype => access_key_secret_key`.

rvsia: I thought we had made this decision in this PR: https://github.com/RedHatInsights/sources-api/pull/310/files#diff-e77d362e01f7845504382af8c7c60f0bbacfab1fbc642e30c2d776928bd942e8R7

lindgrenj6: Updating in #324

PR #315, opened 2021-02-05
Clowder migration

slemrmartin:
**Issue** https://github.com/RedHatInsights/sources-api/issues/318

---

[RHCLOUD-12240](https://issues.redhat.com/browse/RHCLOUD-12240)


slemrmartin:
@lindgrenj6 ready for review. Should be compatible with CI

Gemfile
```diff
@@ -3,6 +3,9 @@ source 'https://rubygems.org'
 plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
+# Parser for Clowder config in ENV['ACG_CONFIG'] path
+gem 'app-common-ruby', :git => 'https://github.com/slemrmartin/app-common-ruby', :branch => 'master', :require => false
```
lindgrenj6, line 7: Not a blocker, but we can probably change this to the insights repo right?

deploy/clowdapp.yaml
```diff
@@ -0,0 +1,201 @@
...
+        public:
+          enabled: true # port 8000 is default in ClowdEnvironment
+      podSpec:
+        image: ${IMAGE}:${IMAGE_TAG}
+        initContainers:
+        - command:
+          - entrypoint
+          args:
+          - seed_database
+          inheritEnv: false
```
lindgrenj6, line 26: Is there a reason we don't want to use this? 

I think we could do `inheritEnv: true` here, and just move `SOURCES_ENV` to the main `podSpec`. that way we don't have duplicated keys.

deploy/clowdapp.yaml
```diff
@@ -0,0 +1,201 @@
...
+      name: sources # looking for rds: sources-db resource in app-interface
+      version: 12
+    kafkaTopics:
+    - topicName: platform.sources.event-stream
+      partitions: 3
+      replicas: 3
+    - topicName: platform.sources.status
+      partitions: 3
+      replicas: 3
+    - topicName: platform.sources.superkey-requests
```
lindgrenj6, line 104: Thank you for including this.

docker-assets/entrypoint
```diff
@@ -4,10 +4,38 @@ function urlescape() {
...
+
+  export WEB_PORT=${CONFIG_VALUES[5]}
+fi
+
 safeuser=$(urlescape ${DATABASE_USER})
 safepass=$(urlescape ${DATABASE_PASSWORD})
 
 export RAILS_ENV=production
-export DATABASE_URL="postgresql://${safeuser}:${safepass}@${DATABASE_HOST}:${DATABASE_PORT}/sources_production?encoding=utf8&pool=5&wait_timeout=5"
+export DATABASE_URL="postgresql://${safeuser}:${safepass}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
```
lindgrenj6, line 39: :tada:

docker-assets/entrypoint
```diff
@@ -4,10 +4,38 @@ function urlescape() {
...
+  export WEB_PORT=3000
+else
+  CONFIG_STRING=`ruby<<EOF
+  require 'json'
+  file = File.read(ENV['ACG_CONFIG'])
+  data = JSON.parse(file)
+  puts data['database']['hostname']
+  puts data['database']['port']
+  puts data['database']['username']
+  puts data['database']['password']
```
lindgrenj6, line 18: I'm not sure we should be `put`'ing the password to STDOUT.

Gemfile
slemrmartin: @lindgrenj6 we want to change it to rubygems, but there is a discussion about the name of gem at this moment

deploy/clowdapp.yaml
slemrmartin: we don't need all ENVs in init container and don't need SOURCES_ENV in the main container, so it's not duplicated.
This seems more clear to me, so is there a reason for `inheritEnv: true`?

docker-assets/entrypoint
slemrmartin: Any idea how to parse the json file better in shell?

lindgrenj6: Oh. I missed that it's just redirecting. I apologize.

deploy/clowdapp.yaml
lindgrenj6: Yeah I'm feeling my review here wasn't in-depth enough, I also forgot that the DB settings are getting read in from clowder now not from the ENV vars. So there is no duplication. I apologize.

lindgrenj6 (approved):

PR #316, opened 2021-02-09
Availability status concern modifications

pkomanek:
Adding a new functionality for availability_status concern. From now there will be some kind of interactions between models after attribute change.

Attribute change:
  * Endpoint:
         - reset realted source status
  * Application:
         - reset realted source status(only if there is no related endpoint)

After discussion we decided to remove availability_status concerne from Authentication model, since the status there is not used anymore.

Based on https://github.com/RedHatInsights/sources-api/issues/317


app/models/authentication.rb
```diff
@@ -25,4 +25,23 @@ def set_source
...
+  end
+
+  def reset_related_applications_status
+    applications.each(&:remove_availability_status!)
+  end
+
+  def reset_related_endpoint_status
+    return unless resource_type != `Endpoint`
+
+    resource.remove_availability_status!
```
lindgrenj6, line 45: Curious, why does this guard clause exist for endpoints but not applications?

app/models/application.rb
```diff
@@ -14,6 +14,17 @@ class Application < ApplicationRecord
 
   validate :source_must_be_compatible
 
+  def remove_availability_status!
+    remove_availability_status
+    save
```
lindgrenj6, line 19: I would use `save!` instead, that way it raises on a failure, I think it fits better with the method name since it ends in `!`

edit: here/endpoint etc.

app/models/source.rb
```diff
@@ -20,8 +20,57 @@ class Source < ApplicationRecord
...
+  end
+
+  def remove_availability_status!(source = nil)
+    remove_availability_status(source)
+    save
+  end
+
+    # testing if the availability_status request code produce the PR test timeout
+
+=begin
```
lindgrenj6, line 45: This isn't going to be committed right? This is just for the PR test?

app/models/concerns/availability_status_concern.rb
```diff
@@ -19,13 +19,22 @@ module AvailabilityStatusConcern
...
+    if respond_to?(:availability_status_error)
+      self.availability_status_error = nil
+    end
+  end
 
-      if respond_to?(:availability_status_error)
-        self.availability_status_error = nil
-      end
+  def remove_availability_status_on_source
+    if availability_status.nil? || changed.include?(:availability_status)
```
lindgrenj6, line 36: So this conditional is a bit odd to me - I think checking to see if the changes include availability status would be enough, since the `||` would fall through to the second conditional almost every time unless it was just created since `nil` is the default. 

If that's the reasoning there should at least be a comment explaining that, or we could have a default on this field since that is what it's doing.

app/models/authentication.rb
```diff
@@ -25,4 +25,23 @@ def set_source
...
+    super
+
+    if availability_status.nil? || changed.include?(:availability_status)
+      reset_related_applications_status
+      reset_related_endpoint_status
+    end
+  end
+
+  def reset_related_applications_status
+    applications.each(&:remove_availability_status!)
```
lindgrenj6, line 39: Shouldn't this be `resource` as well? So we could just have one method that removes the sub-resource's availability_status

app/models/authentication.rb
```diff
@@ -25,4 +25,25 @@ def set_source
                        resource.try(:source_id)
                      end
   end
+
+  def update_status
+    super
+
+    if changed.include?("availability_status")
```
slemrmartin, line 32: this will never happen, because at least operation workers (and I believe neither cost/subswatch) never set this value

app/models/authentication.rb
slemrmartin: this call should be moved above the return (because it's about application_authentications relation)
and the `resource` call should be added

app/models/concerns/availability_status_concern.rb
slemrmartin: as I posted above, this condition will be never met for Authentication, because it's never set

slemrmartin (changes requested):

syncrou:
@lindgrenj6 @slemrmartin it looks like the feedback has been applied.  How is this looking?

lindgrenj6 (approved):

slemrmartin (approved):
:+1: LGTM

PR #324, opened 2021-02-23
Lookup the SuperKey authtype rather than assuming it is just "superkey"

lindgrenj6:
As found in the initial integration by @rvsia and mentioned after merge in #314, instead of setting `:authtype => "superkey"` in the payload it will be better to look up the sourcetype[:authentication] hash and find out which authentication type is the superkey type.

\# TODO:
- [x] update `Source#super_key` method to look up authentication
- [x] update callback to change availability status + error if there is no superkey set + account_authorization in app_creation_workflow
- [x] partition `authentications` array based on new lookup rather than just `:authtype == "superkey"`
- [x] validate on authentication creation that there is only *ONE* superkey authentication per source

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -1,6 +1,4 @@
 class Application < ApplicationRecord
-  SUPERKEY_WORKFLOW = "account_authorization".freeze
-
   include TenancyConcern
   include EventConcern
   include AvailabilityStatusConcern
@@ -40,10 +38,13 @@ class Application < ApplicationRecord
   end
 
   def superkey_workflow
-    return unless source.app_creation_workflow == SUPERKEY_WORKFLOW
+    return unless source.super_key?
 
-    if source.super_key.nil?
-      update!(:availability_status => "unavailable", :availability_status_error => "Superkey Credential Missing")
+    if source.super_key_credential.nil?
+      update!(
+        :availability_status       => "unavailable",
+        :availability_status_error => "The source is missing credentials for account authorization. Please remove the source and try to add it again / open a ticket to solve this issue."
+      )
       return
     end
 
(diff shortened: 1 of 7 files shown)


app/models/application.rb
```diff
@@ -31,6 +31,11 @@ def source_must_be_compatible
   def superkey_workflow
     return unless source.app_creation_workflow == SUPERKEY_WORKFLOW
 
+    if source.super_key.nil?
+      update!(:availability_status => "unavailable", :availability_status_error => "Superkey Credential Missing")
```
rvsia, line 35: Because this error can be shown in the UI, maybe we could make it more user friendly and tell users what to do.

example:
"The source is missing credentials for account authorization. Please remove the source and try to add it again / open a ticket to solve this issue."

app/models/authentication.rb
```diff
@@ -24,4 +25,11 @@ def set_source
                        resource.try(:source_id)
                      end
   end
+
+  def only_one_superkey
+    superkey_authtype = source.source_type.superkey_authtype
+    return unless superkey_authtype && authtype == superkey_authtype
+
+    source.authentications.count { |auth| auth.authtype == superkey_authtype } > 1
```
syncrou, line 33: What if `source.authentications.count { |auth| auth.authtype == superkey_authtype }` was equal to 3 or 5?  Wouldn't this still return true?

Maybe that would be fine - it's just that the method name stipulates `only_one_superkey`.

app/models/application.rb
lindgrenj6: You got it! Updating.

app/models/authentication.rb
lindgrenj6: Updated it to be a bit more readable, using `.any?` rather than `> 1`.

syncrou (approved):
LGTM 👍

Issue #317, opened 2021-02-11
Models interactions with availability_status


pkomanek:
# Description
This issue is focused on sources-api models interactions and handling their availability_status (`Authentication`, `Application`, `Endpoint`, `Source`). We recently added the availability_concern, which removes the availability_status from model after any attribute change(except the IGNORE_LIST*). 
```
  IGNORE_LIST = [
    "availability_status",
    "availability_status_error",
    "last_available_at",
    "last_checked_at",
    "updated_at",
    "name"
  ].freeze
```
We also wanna add some interactions between models. So for example clearing the status from `Endpoint` should force related `Source` to do the same. We also consider to call for availability check after setting the `Source` status to `nil`.

 # Models relations
![relations_state](https://user-images.githubusercontent.com/19405716/107619957-3eb47a80-6c54-11eb-8c72-c492e96cd6ee.jpeg)

# Models interactions(attribute changed)
![status_change](https://user-images.githubusercontent.com/19405716/107622771-981ea880-6c58-11eb-9a7d-0605efbac117.jpg)

@lindgrenj6 @slemrmartin What do you think about these interactions? Also do we want to add a call for the `#availability_check` after updating the status on `Source`(only if status is `nil`)? And What about removing the Models? For example from my point of view removing the `Endpoint` should also force the same interaction as the status_change.

Based on these Jira tickets:
* https://issues.redhat.com/browse/RHCLOUD-11681
* https://issues.redhat.com/browse/RHCLOUD-11552
* https://issues.redhat.com/browse/RHCLOUD-10278

pkomanek:
I would like to split this into three different issues:

1) Interactions between models(after status change to nil)
    * Auth:
         - reset realted source status
         - reset related endpoint status
         - reset status for all related applications
     * Endpoint:
         - reset realted source status
     * App:
         - reset realted source status(only if there is no related endpoint)
2) Request new availability check after Source status is set to nil
3) Modify the availability status check logic to aggregate the status from all related models
    * for example I think that right now if there is no related endpoint, we just check the first app status.

What do you think @lindgrenj6 and @slemrmartin?

slemrmartin:
- At first we should consider if the availability status for Authentication makes sense. I don't see any usage so we could remove these columns
- Other interactions looks good to me
- Deleting of the Endpoint/App/Auth should reset Source's status
- And yes, reset Source#availability_status to `nil` should invoke availability check immediatelly. 

[...]

PR #323, opened 2021-02-22
[WIP]Add after_destroy callback for App, Auth and Endpoint models

pkomanek:
Adding `after_destroy`callback for Authentication/Application/Endpoint with requesting a new `availability_status` check on related source.

Depends on https://github.com/RedHatInsights/sources-api/pull/316.
Based on https://github.com/RedHatInsights/sources-api/issues/317.

PR #325, opened 2021-02-24
Removing status from authentication model

pkomanek:
Adding migration for removing `status` and `status_details` columns from `authentications` table. 
```
    remove_column :authentications, :status,         :string
    remove_column :authentications, :status_details, :string
```
Based on https://github.com/RedHatInsights/sources-api/issues/317

Edit: removing the `status` and `status_details` instead of the `availability_status` related attributes based on the @lindgrenj6 and @rvsia comments.

### DIFF
--- a/db/migrate/20210224151150_remove_status_from_authentication.rb
+++ b/db/migrate/20210224151150_remove_status_from_authentication.rb
@@ -1,0 +1,6 @@
+class RemoveStatusFromAuthentication < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :authentications, :status,         :string
+    remove_column :authentications, :status_details, :string
+  end
+end
--- a/public/doc/openapi-3-v1.0.json
+++ b/public/doc/openapi-3-v1.0.json
@@ -1438,13 +1438,6 @@
           "source_id": {
             "$ref": "#/components/schemas/ID"
           },
-          "status": {
-            "example": "valid",
-            "type": "string"
-          },
-          "status_details": {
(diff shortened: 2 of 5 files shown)


rvsia:
Are we really want to remove it? 🤔 

@lindgrenj6 Where will we store errors for super key authentications?

lindgrenj6:
I agree with @rvsia, remove status and status_details but we need to leave `availability_status` and `availability_status_error` for superkey so we can put a message there if the credential is bad.

app/models/authentication.rb
```diff
@@ -2,7 +2,6 @@ class Authentication < ApplicationRecord
   include PasswordConcern
   include TenancyConcern
   include EventConcern
-  include AvailabilityStatusConcern
```
lindgrenj6, line 5: Shouldn't we leave this in since we have the `availability_status` columns still? It's just the `status/status_details` ones that got removed.

pkomanek:
@lindgrenj6 I cut this PR only to remove the columns from DB and rest of the code will be inside https://github.com/RedHatInsights/sources-api/pull/323/commits.

lindgrenj6:
Can we remove it from the openapi.json files too? 

example here: https://github.com/lindgrenj6/sources-api/blob/d14fa11a824a882e9634e59e37b572b07c6ea31c/public/doc/openapi-3-v3.1.json#L1880-L1886

But I would like it removed from all 4 openapi.json files.

lindgrenj6 (approved):

PR #326, opened 2021-02-24
raise create events when resources are created via /bulk-create

lindgrenj6:
cc @mkanoor

I forgot to include the `#{model}.create` when creating via /bulk_create. This fixes that! Whoops!!


mkanoor:
Looks good, can we get this merged @syncrou we are blocked by it.

slemrmartin (approved):
:+1:

PR #328, opened 2021-03-01
Add new `Bulk.create` event when `/bulk_create` is called.

lindgrenj6:
cc @dccurtis :)

Raising a new event on `/bulk_create`, the older `#{model}.create` messages will still go out for backwards compatibility (mostly for topo services), but there will now also be a new `Bulk.create` message published.

The body of the message is an object of the types + id's that were created. Example:

If one posts a body with:
```
{
    "sources": [(source fields for creation)...],
    "applications": [(app fields for creation)...]
}
```
The message will be:
```
{
    "sources": {"1", "2", "3"},
    "applications": {"4", "5", "6"}
}
```

where the numbers in the message are the id's of what was created. This should make it easier to know which sources/applications/authentications are tied together.

### DIFF
--- a/app/controllers/api/v3x1/bulk_create_controller.rb
+++ b/app/controllers/api/v3x1/bulk_create_controller.rb
@@ -23,6 +23,9 @@ module Api
           end
         end
 
+        # raise a single bulk-create event with the types + array of id's
+        raise_event("Bulk.create", bulk.output.transform_values { |val| val&.map(&:id) }.as_json)
+
         render :status => 201, :json => bulk.output
       end
     end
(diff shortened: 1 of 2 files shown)


lindgrenj6:
cc @dccurtis wrong dcurtis, whoops

slemrmartin (commented):
not sure how the receiver will handle this.
If svc skips all the `{model}.create` events and waits for `Bulk.create` -  if somebody uses simple create for one model, the `Bulk.create` won't come

lindgrenj6:
@slemrmartin yes thats right, but this is what the UI is moving to as well as the rest of the clients. 

They don't want messages that can come out of order, they want the relevant items all packaged together. This is the first step to that, since the users can just process these messages (and subsequent bulk update/delete messages in the future), and topo's stuff can just use the old messages.

slemrmartin:
I understand, but if UI and some services will use bulk create, some other clients using API won't and their POSTs won't be processed by Cost (or whoever will rely _only_ on Bulk.create event)

lindgrenj6:
Closing for now, will probably come back to it later.

PR #332, opened 2021-03-04
Add 'app_creation_workflow' to /bulk_create 'sources' payload

lindgrenj6:
Zvika found this this morning, I forgot to include it in the OpenAPI spec for the bulk create payload.


slemrmartin (approved):
:+1:

PR #336, opened 2021-03-08
Migrate so any dangling Authentications with source_id unset get set

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-12814

A user realized there could be instances where `Authentication#source_id` is unset due to us not running a migration when we initially added the `source_id` column to Authentications.

Note this does not fix the fact that there DEFINITELY are dangling authentications out there - e.g. in CI there are like 200+ that are pointing to an endpoint/application that just doesn't exist.


syncrou (approved):
LGTM 👍 

If we could just have a Jira ticket around this would be great.

lindgrenj6:
@syncrou https://issues.redhat.com/browse/RHCLOUD-12814

PR #341, opened 2021-03-10
Use model SourceType in migration RenameSatelliteInitialValue

lpichler:
This change keeps migrations independent  on `app/models`.


slemrmartin (commented):
@lpichler this is the only migration with model in sources api? 
I mean if it could be changed, then all in one PR

lpichler:
> @lpichler this is the only migration with model in sources api?
> I mean if it could be changed, then all in one PR

yes, this the only migration.

slemrmartin (approved):

PR #344, opened 2021-03-11
Allow sending raise_event only for Source#update

lpichler:
Stop sending `raise_event` for 
`Application#update`, `Endpoint#update` and `Authentication#update` when 
attributes `availability_status` or ` availability_status_error` are present in update request:

```
./api/sources/patch 'applications/1' '{"availability_status":"available"}'
./api/sources/patch 'authentications/1' '{"availability_status":"available"}'
./api/sources/patch 'endpoints/1' '{"availability_status":"available"}'
```

---
https://issues.redhat.com/browse/RHCLOUD-12860

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -74,6 +74,10 @@ class ApplicationController < ActionController::API
     send("api_#{version}_#{endpoint}_url", instance.id)
   end
 
+  def raise_event_if(ignore_raise_event, event, payload)
+    raise_event(event, payload) unless ignore_raise_event
+  end
+
   def raise_event(event, payload)
     headers = Insights::API::Common::Request.current_forwardable
     logger.debug("publishing message to topic \"platform.sources.event-stream\"...")
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -75,14 +75,14 @@ class ApplicationController < ActionController::API
   end
 
   def raise_event_if(ignore_raise_event, event, payload)
-    raise_event(event, payload) unless ignore_raise_event
+    return if ignore_raise_event
+
+    headers = Insights::API::Common::Request.current_forwardable
+    Sources::Api::Events.raise_event_with_logging(event, payload, headers)
   end
 
   def raise_event(event, payload)
-    headers = Insights::API::Common::Request.current_forwardable
-    logger.debug("publishing message to topic \"platform.sources.event-stream\"...")
-    Sources::Api::Events.raise_event(event, payload, headers)
-    logger.debug("publishing message to topic \"platform.sources.event-stream\"...Complete")
+    raise_event_if(false, event, payload)
   end
 
   def params_for_create
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -18,12 +18,6 @@ class Application < ApplicationRecord
   after_create :superkey_workflow
   after_destroy :superkey_workflow
 
-  IGNORE_RAISE_EVENT_ATTRIBUTES_LIST = %i[availability_status availability_status_error].freeze
-
-  def ignore_raise_event_for?(attributes)
-    (IGNORE_RAISE_EVENT_ATTRIBUTES_LIST & attributes.map(&:to_sym)).present?
-  end
-
   def remove_availability_status!
     remove_availability_status
     save!
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -18,6 +18,12 @@ class Application < ApplicationRecord
   after_create :superkey_workflow
   after_destroy :superkey_workflow
 
+  IGNORE_RAISE_EVENT_ATTRIBUTES_LIST = %i[availability_status availability_status_error].freeze
+
+  def ignore_raise_event_for?(attributes)
+    (IGNORE_RAISE_EVENT_ATTRIBUTES_LIST & attributes.map(&:to_sym)).present?
+  end
+
   def remove_availability_status!
     remove_availability_status
     save!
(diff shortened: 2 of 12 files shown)


app/models/application.rb
```diff
@@ -18,6 +18,12 @@ class Application < ApplicationRecord
   after_create :superkey_workflow
   after_destroy :superkey_workflow
 
+  IGNORE_RAISE_EVENT_ATTRIBUTES_LIST = %w[availability_status availability_status_error].freeze
```
syncrou, line 21: Do we ever see a need to export these out to Environment variables ( potentially different for differrent envs?)  Or is a constant going to be fine

syncrou (commented):
Looks great! 1 small question.

slemrmartin (commented):
there is also raise_event in lib/availability_status_listener

app/models/application.rb
lpichler: it looks that it is enough to have those attributes just in constant as raising event `Application#update` (after updating  `availability_status` or `availability_status_error`) doesn't any consumer regardless of environments.

@slemrmartin can you confirm it, please ?  Thanks!

lpichler:
> there is also raise_event in lib/availability_status_listener

I see, PR is updated to also cover availability_status_listener.

app/controllers/application_controller.rb
```diff
@@ -74,11 +74,15 @@ def instance_link(instance)
     send("api_#{version}_#{endpoint}_url", instance.id)
   end
 
-  def raise_event(event, payload)
+  def raise_event_if(ignore_raise_event, event, payload)
+    return if ignore_raise_event
```
syncrou, line 78: I like the name of this method.

syncrou (commented):
LGTM ( codewise ) I'll defer to @slemrmartin @lindgrenj6 for any missing details

syncrou:
@lpichler can we pull the [WIP] out of the name?  

( only nit )

lindgrenj6 (approved):
LGTM, I would like to refactor #345 to use these new methods.

lpichler:
Thanks for the review @syncrou @lindgrenj6! 

> @lpichler can we pull the [WIP] out of the name?

Yes and it is finished  - I added last in thing (as we discussed with @slemrmartin) 
that we need `model.update` only for `Source` and 
it was added in last [commit](https://github.com/RedHatInsights/sources-api/pull/344/commits/b608ea6df5b121057f067981f13cdd821e11901b).

Description changed to fit current behaviour.

lindgrenj6:
@slemrmartin can you look at this? Looking to get it in so I can refactor my superkey PR.

app/models/concerns/event_concern.rb
```diff
@@ -2,13 +2,36 @@ module EventConcern
...
+    "Application"    => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST,
+    "Authentication" => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST,
+    "Endpoint"       => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST
+  }.freeze
+
+  def ignore_raise_event_for?(attributes)
+    ignore_attribute_list = IGNORE_RAISE_EVENT_LIST[self.class.name]
+    return false unless ignore_attribute_list
+
+    (IGNORE_RAISE_EVENT_ATTRIBUTES_LIST & attributes.map(&:to_sym)).present?
```
slemrmartin, line 20: I think the event should be ignored if there is _only_ one or more attributes from the ignore list. 
@lindgrenj6 what do you think?

lib/availability_status_listener.rb
```diff
@@ -52,7 +52,7 @@ def process_event(event)
     options[:last_available_at] = options[:last_checked_at] if options[:availability_status] == 'available'
 
     object.update!(options)
-    raise_event("#{model_class}.update", object.as_json, event.headers)
+    object.raise_event_for_update(options.keys)
```
slemrmartin, line 55: I'm not sure if the headers will be present (they are assigned to the Insights::Api::Common::Request by Controller, right?)

app/models/concerns/event_concern.rb
lindgrenj6: I'm on the other side, if the application (say cost) PATCH's back into us and updates `extra` at the same time as `availability_status_error` they still are not going to want the update since they are the ones that updated it. They're aware of any changes that go into the application when they're updating the availability status. 

The `availability_status` fields will most likely only ever be updated by the applications themselves so it's safe to ignore them even if they aren't the only fields IMO.

lib/availability_status_listener.rb
lindgrenj6: We pass the headers through the kafka event, so yes this won't work with the new `safe_headers` method. 

Not sure how we can get around it other than having the listener just do the manual raise with an if around it or just having a default parameter.

app/models/concerns/event_concern.rb
slemrmartin: ok, it makes sense

lib/availability_status_listener.rb
slemrmartin: have the `safe_headers` as a default value for 2nd param of `raise_event_for_update` seems to be fine to me

app/models/application.rb
slemrmartin: Constant is enough. I don't think we'll need different values per env

lib/availability_status_listener.rb
lpichler: @slemrmartin Good catch! @lindgrenj6 thanks for advice 👍 

Now fixed 👍

lpichler:
I updated PR according to last suggestions 👍 

thanks!

slemrmartin (approved):
:+1: good job!

PR #345, opened 2021-03-15
Only raise Application.create events for Superkey sources when complete

lindgrenj6:
During the integration process for superkey, I'm finding that ordering of messages is the biggest problem faced.

So I'm reworking to only raise the `Application.create` event once the superkey worker posts back. It's complicated logic, but it's necessary to make sure that the events go out in the right order:

1. `Source.create`
2. `Application.create`
3. `Authentication.create`
4. `ApplicationAuthentication.create`

\# TODO:
- [x] tests

### DIFF
--- a/app/controllers/api/v1/applications_controller.rb
+++ b/app/controllers/api/v1/applications_controller.rb
@@ -26,7 +26,7 @@ module Api
         # values for the source, but only the first time after the worker processes the message,
         # from there on we raise the normal update.
         first_time_superkey = (application.source.super_key? && params.try(:[], "extra")&.key?("_superkey"))
-        raise_event_unless(first_time_superkey, "Application.create", application.as_json)
+        raise_event_if(first_time_superkey, "Application.create", application.as_json)
 
         # appending the extra keys in case of _superkey being updated.
         application.raise_event_for_update(params_for_update.keys + params.fetch("extra", {}).keys)
--- a/app/controllers/api/v1/applications_controller.rb
+++ b/app/controllers/api/v1/applications_controller.rb
@@ -4,14 +4,34 @@ module Api
       include Api::V1::Mixins::DestroyMixin
       include Api::V1::Mixins::IndexMixin
       include Api::V1::Mixins::ShowMixin
-      include Api::V1::Mixins::UpdateMixin
 
       def create
         application = Application.new(params_for_create).tap { |app| authorize(app) }
         application.save!
 
-        raise_event("#{model}.create", application.as_json)
-        render :json => application, :status => :created, :location => instance_link(application)
+        # we do not want to raise the create event since the application has
+        # not been processed by the superkey worker.
+        raise_event_if(!application.source.super_key?, "Application.create", application.as_json)
+
+        render :json => application, :status => 201, :location => instance_link(application)
+      end
+
+      def update
+        application = Application.find(params.require(:id))
+        authorize(application)
+
+        application.update!(params_for_update)
+
+        # Here we're raising the create event after the worker has filled in the
+        # values for the source, but only the first time after the worker processes the message,
+        # from there on we raise the normal update.
+        first_time_superkey = (application.source.super_key? && params.try(:[], "extra")&.key?("_superkey"))
+        raise_event_unless(first_time_superkey, "Application.create", application.as_json)
+
+        # appending the extra keys in case of _superkey being updated.
+        application.raise_event_for_update(params_for_update.keys + params.fetch("extra", {}).keys)
+
+        head :no_content
       end
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -74,19 +74,15 @@ class ApplicationController < ActionController::API
     send("api_#{version}_#{endpoint}_url", instance.id)
   end
 
-  def raise_event_if(ignore_raise_event, event, payload)
-    return if ignore_raise_event
+  def raise_event_if(raise_event_allowed, event, payload)
+    return unless raise_event_allowed
 
     headers = Insights::API::Common::Request.current_forwardable
     Sources::Api::Events.raise_event_with_logging(event, payload, headers)
   end
 
-  def raise_event_unless(ignore_raise_event, event, payload)
-    raise_event_if(!ignore_raise_event, event, payload)
-  end
-
   def raise_event(event, payload)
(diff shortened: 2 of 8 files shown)


lindgrenj6:
cc @syncrou

app/controllers/api/v1/applications_controller.rb
```diff
@@ -4,15 +4,35 @@ class ApplicationsController < ApplicationController
...
-      include Api::V1::Mixins::UpdateMixin
 
       def create
         application = Application.new(params_for_create).tap { |app| authorize(app) }
         application.save!
 
-        raise_event("#{model}.create", application.as_json)
+        # we do not want to raise the create event since the application has
+        # not been processed by the superkey worker.
+        raise_event("Application.create", application.as_json) unless application.source.super_key?
```
syncrou, line 14: I'm totally outside this codebase, but can any of the `Event` concern methods be used here? `app/models/concerns/event_concern.rb` so we could hide some of the conditional logic within a `raise_event` method?

syncrou (commented):
Behavior looks good. My only nit, can we hide the conditionals for `raise_event` in the `event_concern` at all like what we're doing here: https://github.com/RedHatInsights/sources-api/pull/344/files

I realize I've been out of this code for a while so feel free to let me know if this is not a workable idea.

app/controllers/api/v1/applications_controller.rb
lindgrenj6: I really like the look of #344, so I'll wait on that one and refactor this to use the `raise_event_if` pattern.

lindgrenj6:
So it appears some of the logic is backwards for some conditions, e.g. I added a `raise_event_unless` method that inverses the check, but it's still hard to read inverse checks.

examples:
1.
```
raise_event_if(!application.source.super_key?, "Application.create", ...)
```
This reads fine, we're raising if the application is not a super key.


2.
```
first_time_superkey = (stuff resolving to true, indicating that this is the superkey application update)
raise_event_unless(first_time_superkey, "Application.create", application.as_json)
```
This one seems...off, `raise_event_if(first_time_superkey,..)` doesn't do the right behavior, so I added `raise_event_unless`, but that just reads backwards.


3.
```
raise_event_unless(!app.source.super_key?, "Application.create", app.as_json)
```
This one reads completely backwards. But `raise_event_if(!app.source.super_key?, ...)` doesn't work, and `raise_event_unless(app.source.super_key?, ...)` doesn't work either. 
So we have the double negative and it works. 

I'm not sure why it is this way, I'll keep playing with it some more.

slemrmartin:
What about this form:
```
not_superkey = !app.source.superkey?
raise_event_if(not_superkey, ...)
```

lindgrenj6:
@slemrmartin I added a commit that just reverses the `raise_event_if` logic.

Now it reads like this:
`raise_event_if(truthy_val, message, values, headers)`

This makes the events read a little better imo since if is usually checking for something that resolves to `true`. 

Let me know what you think. cc @lpichler

app/models/concerns/event_concern.rb
```diff
@@ -8,16 +8,16 @@ module EventConcern
...
-    "Application"    => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST,
+    "Application"    => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST + %i[_superkey],
     "Authentication" => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST,
     "Endpoint"       => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST
   }.freeze
 
   def ignore_raise_event_for?(attributes)
     ignore_attribute_list = IGNORE_RAISE_EVENT_LIST[self.class.name]
-    return false unless ignore_attribute_list
+    return true unless ignore_attribute_list
```
lpichler, line 18: @lindgrenj6 Should not we neg also name of this method in this case?
something like `ignore_raise_event_for` -> `allow_raise_event_or`.
Original intention was - what is in `IGNORE_RAISE_EVENT_LIST` it should be ignored
so If model is not present  in `IGNORE_RAISE_EVENT_LIST` we should not ignore raising event and it means we should raise event.

app/models/concerns/event_concern.rb
```diff
@@ -8,16 +8,16 @@ module EventConcern
...
 
-    (IGNORE_RAISE_EVENT_ATTRIBUTES_LIST & attributes.map(&:to_sym)).present?
+    (ignore_attribute_list & attributes.map(&:to_sym)).empty?
```
lpichler, line 20: thanks for `IGNORE_RAISE_EVENT_LIST` -> `ignore_attribute_list` 👍

app/controllers/application_controller.rb
```diff
@@ -74,15 +74,15 @@ def instance_link(instance)
     send("api_#{version}_#{endpoint}_url", instance.id)
   end
 
-  def raise_event_if(ignore_raise_event, event, payload)
-    return if ignore_raise_event
+  def raise_event_if(raise_event, event, payload)
+    return unless raise_event
```
lpichler, line 78: yes `ignore_raise_event` was not great but what about to name this more like `raise_event_allowed` or  `raise_event_expression` or `raise_event_condition`
or something like that express that it is condition

lpichler:
> @slemrmartin I added a commit that just reverses the `raise_event_if` logic.
> 
> Now it reads like this:
> `raise_event_if(truthy_val, message, values, headers)`
> 
> This makes the events read a little better imo since if is usually checking for something that resolves to `true`.
> 
> Let me know what you think. cc @lpichler

I added some suggestions but they make sense to you but yes naming is hard. Otherwise looks good 👍 
thanks for reducing "if"s.

app/models/concerns/event_concern.rb
lindgrenj6: Absolutely, I renamed the method to account for the reversed behavior. Sorry I missed that.

app/controllers/application_controller.rb
lindgrenj6: Changed to `raise_event_allowed` :)

syncrou (approved):
LGTM 👍 - with the current tacit approval from @lpichler @slemrmartin I am merging this so we can get input from the app teams.

If either of you feel this should have additional changes please raise with @lindgrenj6 and we'll get them in followup PRs.

PR #347, opened 2021-03-18
Add more validations when creating and destroying SuperKey Authentications

lindgrenj6:
Fixing: 
- https://issues.redhat.com/browse/RHCLOUD-12897
    This validation specifies that an Authentication can _only_ be attached to a Source if the `Source#app_creation_workflow` is `Source::SUPERKEY_WORKFLOW`

- https://issues.redhat.com/browse/RHCLOUD-13031
    This validation makes it so the user has to disable all of the applications associated with the source before deleting the source, this is just due to the fact that if the Source gets deleted (thus cascade deleting the authentication) there is potential for dangling resources in AWS. 

    This validation ensures that cannot happen, since the superkey worker will have plenty of time to clean up the resources before the Source + SuperKey Auth get deleted in the API

\# TODO: 
- [x] specs

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -22,6 +22,8 @@ class Source < ApplicationRecord
   validates :name, :presence => true, :allow_blank => false,
             :uniqueness => { :scope => :tenant_id }
 
+  before_destroy :validate_no_applications, :if => -> { super_key? }, :prepend => true
+
   def default_endpoint
     default = endpoints.detect(&:default)
     default || endpoints.build(:default => true, :tenant => tenant)
@@ -46,5 +48,9 @@ class Source < ApplicationRecord
   def remove_availability_status!(source = nil)
     remove_availability_status(source)
     save!
+  end
+
+  def validate_no_applications
+    raise(ActiveRecord::RecordNotDestroyed, "Applications must be removed before destroying parent source") if applications.any?
   end
 end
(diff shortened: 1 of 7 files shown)


lpichler (approved):
👍

app/models/source.rb
```diff
@@ -47,4 +49,8 @@ def remove_availability_status!(source = nil)
     remove_availability_status(source)
     save!
   end
+
+  def validate_no_applications
+    raise(ActiveRecord::RecordNotDestroyed, "Applications must be removed before destroying parent source") if applications.any?
```
slemrmartin, line 54: isn't it redundant? Postgres includes Foreign key on applications:sources_id so it's deleted by database

lindgrenj6: It is automatically deleted by cascade yes - this is just to prevent a potential race condition where a Superkey Source is deleted and if there are any application resources there could be a case where the cascade delete happens too fast and the worker can't pick up the credentials in order to clean up the resources in AWS. 

This hasn't bit us yet, but that's only because the worker basically picks up the message and gets the credential right away. If there is kafka lag or something else that makes it take longer than a split second it would fail to delete the resources.

slemrmartin: Ah, get it

slemrmartin (approved):

PR #349, opened 2021-03-23
Store the headers from the initial Application create operation to send when the superkey worker fills in the application

lindgrenj6:
So it turns out the final issue (for cost at least) is that they key off of the headers during create, thats why they aren't able to process the messages anymore since the superkey worker is doing the posting. This is an initial implementation storing stuff on the sources api side. 

I am going to look into sending the headers to the worker and then using it from there so this isn't necessary.


lindgrenj6:
Doing the work in #350 instead, rather than this convoluted way. Closing.

PR #350, opened 2021-03-23
Pass along forwardable headers to superkey worker

lindgrenj6:
Need to pass along the headers in order to make the request back to sources correctly (dummy won't work for all customers)

Needed for https://github.com/RedHatInsights/sources-superkey-worker/pull/22 to work.


syncrou (approved):
LGTM 👍

syncrou:
Merging without 🟢 since it seems the app sre bot will never okay this...

PR #353, opened 2021-03-29
Add AWS acc number for subwatch in stage/qa

rvsia:
cc @slemrmartin @lindgrenj6

### DIFF
--- a/db/seeds/app_metadata.yml
+++ b/db/seeds/app_metadata.yml
@@ -14,6 +14,8 @@ ci:
 qa:
   "/insights/platform/cost-management":
     gcp_service_account: test-billing-service-account@cloud-billing-292519.iam.gserviceaccount.com
+  "/insights/platform/cloud-meter":
+    aws_wizard_account_number: "372779871274"
 stage:
   "/insights/platform/cost-management":
     gcp_service_account: billing-export@red-hat-cost-management-stage.iam.gserviceaccount.com
@@ -22,3 +24,5 @@ stage:
 prod:
   "/insights/platform/cost-management":
     gcp_service_account: billing-export@red-hat-cost-management.iam.gserviceaccount.com
+  "/insights/platform/cloud-meter":
+    aws_wizard_account_number: "998366406740"


infinitewarp:
# ⚠️ This is bad and will break Cloud Meter in the future! ⚠️ 

Our AWS account IDs should _never_ be hard-coded and are subject to change. When you need to get the Cloud Meter AWS account ID, you can retrieve it from our sysconfig endpoint, which I'm certain is what the Sources UI already does (unless this recently regressed).

lindgrenj6:
@infinitewarp are there docs on how to request this? I'm going to need to update the worker to fetch it dynamically since the UI isn't doing it.

syncrou:
@infinitewarp yes, exactly what @lindgrenj6 is asking.  The only reason static ids are being used are that we haven't been shown how to dynamically look them up. The static values are what were handed to our team. Once we get docs we will make the appropriate changes.

cc - @rvsia

infinitewarp:
Here's an example from our docs. (search for "sysconfig") https://github.com/cloudigrade/cloudigrade/blob/master/docs/rest-api-examples.rst#retrieve-current-publicly-viewable-system-configuration

Observe that a top-level key in the returned JSON is "aws_account_id".

@rvsia should also be able to demonstrate because I believe he implemented the call from the Sources UI.

infinitewarp:
> The static values are what were handed to our team.

@syncrou Who gave these values to you? I only learned about this change through an indirect message from a QE on my team.

lindgrenj6:
@infinitewarp admittedly I just yoinked them from the UI. Cost's acct number is static, so I assumed this one would be good for now at least. 

How often are these going to change?

syncrou:
@infinitewarp as well, I thought the uI was still using static values. So that's on me.

infinitewarp:
They're unlikely to change _frequently_ in stage and prod, but we did just have to make an account change a few months ago, and they're likely to change more frequently for the lower-tier environments, especially ephemeral environments.

syncrou:
@infinitewarp ahh. Ok. So "if" the `account_id` does change on Prod / Stage that would mean all existing sources set with the previous `account_id` would be broken.  Is that expected behavior on your end?

The dynamic nature of account ids used in ephemeral environments makes a lot of sense, so I understand how static values there could cause issues.

infinitewarp:
Any customer with an existing source for Cloud Meter would need to update their AWS IAM Role to include the new backend AWS account ID in its list of trusted entities, but I don't think that would have any effect on the objects in the Sources service itself. The Role's ARN would be unchanged. I'm not very familiar with the current behavior or inner workings of the new super key functionality, but I suppose it could update or would be responsible for updating the Role on the customer's behalf if/when the value changes. That's total speculation on my part, though.

It's _unlikely_ that we'll change our AWS account ID in prod, at least not without alerting customers first since it would require them taking action, but stage and everywhere else are available to change at any time since their data is not business-critical. To be clear, it was stage that changed a few months ago.

Also, and this is projecting somewhat far into the future, if/when we eventually support AWS GovCloud or AWS China, those service "regions" will both have separate AWS account IDs for each environment which will also be dynamic. Just something to be aware of.

syncrou (approved):
LGTM 👍 
For Stage / Prod this solution allows us to move forward.

There is a plan to allow for dynamic values for Ephemeral environments but that will be handled in a follow up.

PR #354, opened 2021-03-29
[WIP] Fetch Subscription Watch info dynamically when running superkey workflow

lindgrenj6:
As discussed in #353, we should be fetching account number + policy JSON at runtime rather than going off of static values, this PR is that work.

\# TODO: 
- [x] Service class to fetch data 
- [x] Key off of application type and whether we should fetch it dynamically or go off of `AppMetaData`

cc @infinitewarp

### DIFF
--- a/lib/sources/subscription_watch_info.rb
+++ b/lib/sources/subscription_watch_info.rb
@@ -1,0 +1,28 @@
+module Sources
+  class SubscriptionWatchInfo
+    X_RH_ID = Base64.strict_encode64({:identity => {:account_number => "12345", :user => {:is_org_admin => true}}}.to_json).freeze
+
+    def self.data
+      @data ||= begin
+        cloudigrade_url = URI.parse(ENV["CLOUD_METER_AVAILABILITY_CHECK_URL"])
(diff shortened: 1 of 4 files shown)


lib/sources/subscription_watch_info.rb
```diff
@@ -0,0 +1,26 @@
+module Sources
+  class SubscriptionWatchInfo
+    XRHID = Base64.strict_encode64({:identity => {:account_number => "12345", :user => {:is_org_admin => true}}}.to_json).freeze
```
lpichler, line 3: `account_number` should be hardcoded ?

lindgrenj6: Yeah this is just a dummy header to fetch from their Internal API, the acct # doesn't matter.  

Might move it out to a svc class haven't decided yet.

lpichler: 👍 thanks

lindgrenj6:
Closing this due to doing it on the superkey side now here: https://github.com/RedHatInsights/sources-superkey-worker/pull/42

PR #355, opened 2021-03-29
add cost account for app_meta_data for all envs

lindgrenj6:


slemrmartin:
@lindgrenj6 is this ready to merge or there is the same issue like in #353 ?

lindgrenj6:
this one is ready to go, 353 can go as well for now. if we need to change to dynamic fetching we can just pull them out then (for swatch).

syncrou (approved):
LGTM pushing to fix Stage. 👍

PR #366, opened 2021-03-30
Add Records.(create|update) messaging

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-12199

Creating the new `SourcesInstance.operation` messages, name subject to change. They will be raise on bulk_create events as well as update/~deletions~.

I will do delete in a follow-up PR.

### DIFF
--- a/app/models/concerns/event_concern.rb
+++ b/app/models/concerns/event_concern.rb
@@ -3,6 +3,56 @@ module EventConcern
 
   included do
     after_destroy :raise_event_for_destroy, :prepend => true
+
+    def bulk_message
+      payload = case self.class.to_s
+                when "Endpoint", "Application"
+                  {
+                    :source                      => source,
+                    :applications                => source.applications,
+                    :endpoints                   => source.endpoints,
+                    :authentications             => authentications,
+                    :application_authentications => ApplicationAuthentication.where(:authentication => authentications)
+                  }
+                when "Authentication"
+                  {
+                    :source                      => resource.source,
+                    :applications                => resource.source.applications,
+                    :endpoints                   => resource.source.endpoints,
+                    :authentications             => resource.authentications,
+                    :application_authentications => ApplicationAuthentication.where(:authentication => resource.authentications)
+                  }
+                when "ApplicationAuthentication"
+                  {
+                    :source                      => application.source,
+                    :applications                => application.source.applications,
+                    :endpoints                   => application.source.endpoints,
+                    :authentications             => application.authentications,
+                    :application_authentications => ApplicationAuthentication.where(:authentication => application.authentications)
+                  }
+                when "Source"
+                  {
+                    :source                      => self,
+                    :applications                => applications,
+                    :endpoints                   => endpoints,
+                    :authentications             => authentications || [] - super_key_credential,
+                    :application_authentications => ApplicationAuthentication.where(:application => applications)
+                  }
+                end
+
+      payload.transform_values(&:as_json)
+    end
+
+    def update_message(attributes)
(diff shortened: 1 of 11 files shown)


slemrmartin (commented):
LGTM, just a few things:
- Why only create and update? (no destroy)
- CRCApp as a name is confusing, it should describe the source of data. Maybe `SourcesAPI.(create|update)` ... or something better

lindgrenj6:
>Why only create and update? (no destroy)

I wanted to just do create + update for now - will follow-up with destroy in a separate PR (since it is a separate ticket as well).

>CRCApp as a name is confusing, it should describe the source of data. Maybe SourcesAPI.(create|update) ... or something better

Yes I am having trouble with the naming as well. I was thinking `CRCApp` = `Cloud dot Redhat dot Com App` since thats basically what it is for the whole application. `SourcesAPIApplication` would work too. hmm. Any other ideas?

slemrmartin:
`SourcesAPIApplication` sounds better

slemrmartin:
or `SourcesAPIService` - the "Application" can be interpreted as a model name

lindgrenj6:
Chewing on this some more - how about `SourcesInstance`, it's an instance in Sources. The only thing I don't like about `SourcesAPIService` is it really isn't creating/updating/deleting the service.

slemrmartin:
hmm, naming is the hardest challenge of programming :laughing: 
I'm still not satisfied, sounds like instance of the Source model. 
Next proposals: 
`SourcesData`, `SourcesAPIData`,  `SourcesRecords`, `SourcesAPIRecords`

@syncrou can you help us? :)

syncrou:
My idea may be equal to "too many cooks in the kitchen" but here goes.

How about dropping `Sources` from the name of the class:  `Data`, `Record`

I don't think the class name should be plural since that goes against common Rails practices.  We already know the app is an API and that the base class ( or god object ) is a `Source` so potentially no need to include that in the class name either.

slemrmartin:
It's true these data are pushed to the topic `platform.sources.event-stream` so everyone knows it's coming from Sources. 
But I'm for plural, because it doesn't describe class name, but set of class names, so `Data` or `Records` sounds good to me

lindgrenj6:
Ok so after letting this marinate for a week I went with `Records`. lmk what you guys think.

slemrmartin (approved):
:+1: cool, I'm ok with `Records`

app/models/concerns/event_concern.rb
```diff
@@ -3,6 +3,56 @@ module EventConcern
...
+
+    def bulk_message
+      payload = case self.class.to_s
+                when "Endpoint", "Application"
+                  {
+                    :source                      => source,
+                    :applications                => source.applications,
+                    :endpoints                   => source.endpoints,
+                    :authentications             => authentications,
+                    :application_authentications => authentications.flat_map(&:application_authentications)
```
lpichler, line 15: @lindgrenj6 

it looks that those queries(and similar queries bellow) produces N+1 queries. (I am not sure how much it is problem regard to count of records)

maybe can be improved with just 2 queries: 

```
ApplicationAuthentication.includes(:authentication).where(:authentications => {:id => authentications.ids})
```

what do you think about it ?

lindgrenj6: Great catch, I didn't think of that. I think I was able to replicate it with what I changed it to - rails automatically selects by id when passing in an array of AR objects. 

LMK what you think!

lpichler: I am sorry for delay, that's right and this looks good, thanks!

lpichler (approved):

PR #368, opened 2021-03-31
only update availability status on superkey applications when new_record?

lindgrenj6:
This appears to be what is causing the error:
```
{
    "errors": [
        {
            "detail": "RuntimeError: Can't modify frozen hash",
            "status": "400"
        }
    ]
}
```

apparently trying to update a subresource's availability status wasn't a good idea on a destroy operation, classic. :facepalm:

cc @zsadeh

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -41,10 +41,16 @@ class Application < ApplicationRecord
     return unless source.super_key?
 
     if source.super_key_credential.nil?
-      update!(
-        :availability_status       => "unavailable",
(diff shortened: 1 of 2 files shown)


syncrou (commented):
@lindgrenj6 should we have specs around this change?

lindgrenj6:
@syncrou great idea, I added some. Not sure why I missed that other than pure laziness!

syncrou (approved):
LGTM 👍

PR #369, opened 2021-04-01
Handle SuperKey deletions asynchronously so the credentials do not get deleted before the AWS resources

lindgrenj6:
So lets try this again. 

~~Adding the validation without alerting anyone turns out to have been a bad idea (oops!), but this validation still needs to be in place so this is paging @rvsia and @zsadeh letting you know that if a DELETE request happens to a superkey source *and* it has applications attached, the delete request will return a 400.~~

~~IQE will have to move to deleting applications first (only for superkey applications) via the API, the UI should just handle the error message coming back, forcing the user to un-check the applications then the delete will work.~~

---

Turns out the biggest issue (delete operation returning the normal 204 even though it wasn't deleted) is all caused by the DestroyMixin not raising if the operation failed. Not sure why that wasn't in place, but the second commit fixes that here. 

EDIT: 5/12/21
Ok this is mostly done, I have the deletion of a source kicking off a job to do so - only problem now is not queuing the job twice for the superkey worker. almost there!

### DIFF
--- a/app/controllers/api/v3x1/sources_controller.rb
+++ b/app/controllers/api/v3x1/sources_controller.rb
@@ -1,0 +1,17 @@
+module Api
+  module V3x1
+    class SourcesController < Api::V3x0::SourcesController
+      def destroy
+        source = Source.find(params.require(:id)).tap { |s| authorize(s) }
+
+        if source.super_key?
+          SuperkeyDeleteJob.perform_later(source, Insights::API::Common::Request.current_forwardable)
+          head :accepted
+        else
+          source.destroy!
+          head :no_content
+        end
+      end
+    end
+  end
+end
--- a/app/jobs/async_delete_job.rb
+++ b/app/jobs/async_delete_job.rb
@@ -1,0 +1,11 @@
+class AsyncDeleteJob < ApplicationJob
+  queue_as :default
+
+  def perform(instance, headers)
+    Sidekiq.logger.info("Destroying #{instance.class} #{instance.id}...")
+
+    Insights::API::Common::Request.with_request(:original_url => "noop", :headers => headers) do
+      instance.destroy!
+    end
+  end
+end
--- a/app/jobs/superkey_delete_job.rb
+++ b/app/jobs/superkey_delete_job.rb
@@ -1,0 +1,25 @@
+class SuperkeyDeleteJob < ApplicationJob
+  queue_as :default
+
+  def perform(source, headers)
+    Insights::API::Common::Request.with_request(:original_url => "noop", :headers => headers) do
+      source.applications.each do |app|
+        sk = Sources::SuperKey.new(
+          :provider    => source.source_type.name,
+          :source_id   => source.id,
+          :application => app
+        )
+
+        app.discard
+        sk.teardown
+
+        # store a key in redis that expires in 30 seconds, that way we don't
+        # double-enqueue a delete job for the superkey resources.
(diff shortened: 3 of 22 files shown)


slemrmartin (commented):
can you pls explain again when exactly it causes an error without validation?
I'm wondering -  if we can't rely on db cascade delete - if we can destroy apps in before_destroy callback and wait for it in the backend

lindgrenj6:
@slemrmartin thinking through this over the weekend - it sounds like we need some sort of job queuing system, so I think I'm going to bring in [sidekiq](https://github.com/mperham/sidekiq) to the sources deployment. 

That way the deletion of a superkey source can go like this:
1. soft-delete the Source model (will also tie in with the source pause work, so thats a plus)
2. send job to sidekiq to delete the superkey source
3. return 204

Then the job will send the delete request to the superkey worker + wait for completion. On completion it will delete the source cascade like normal. Having the ability to queue background jobs will be very nice for superkey things + maintenance tasks as we need them in the future.

slemrmartin:
@lindgrenj6 I used to use resque in rails 3, I don't know what is better. Anyway I was thinking about the same in the context of the POST requests to apps with availablity checks

slemrmartin:
@lindgrenj6 with the pausing and sidekiq merged, are you going to update this PR?

lindgrenj6:
@slemrmartin yes I am working on it.

app/controllers/api/v3x1/sources_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Api
+  module V3x1
+    class SourcesController < Api::V3x0::SourcesController
+      def destroy
+        source = Source.find(params.require(:id)).tap { |s| authorize(s) }
+
+        if source.super_key?
```
slemrmartin, line 7: it's not a subject of this PR, but I can image more descriptive method name, i.e. `source.has_super_key?` ot `source.super_key_auth?`

app/jobs/async_delete_job.rb
```diff
@@ -0,0 +1,11 @@
+class AsyncDeleteJob < ApplicationJob
+  queue_as :default
+
+  def perform(instance, headers)
+    Sidekiq.logger.info("Destroying #{instance.class} #{instance.id}...")
+
+    Insights::API::Common::Request.with_request(:original_url => "noop", :headers => headers) do
```
slemrmartin, line 7: is it working without the parameter - `do |_request| ... end` ?

app/jobs/superkey_delete_job.rb
```diff
@@ -0,0 +1,25 @@
...
+  def perform(source, headers)
+    Insights::API::Common::Request.with_request(:original_url => "noop", :headers => headers) do
+      source.applications.each do |app|
+        sk = Sources::SuperKey.new(
+          :provider    => source.source_type.name,
+          :source_id   => source.id,
+          :application => app
+        )
+
+        app.discard
```
slemrmartin, line 13: does it have some impact at this moment?

app/jobs/superkey_delete_job.rb
```diff
@@ -0,0 +1,25 @@
...
+        sk.teardown
+
+        # store a key in redis that expires in 30 seconds, that way we don't
+        # double-enqueue a delete job for the superkey resources.
+        Redis.current.setex("application_#{app.id}_delete_queued", 30, true)
```
slemrmartin, line 18: :+1:

slemrmartin (approved):
:+1: Great job!

PR #375, opened 2021-04-08
Re-enable Application/Endpoint availability status update messages

lindgrenj6:
Subscription Watch needs the `Application.update` messages regarding availability status. We should have validated that no one else needed them before turning them off. 

cc @syncrou

### DIFF
--- a/app/models/concerns/event_concern.rb
+++ b/app/models/concerns/event_concern.rb
@@ -7,10 +7,11 @@ module EventConcern
 
   IGNORE_RAISE_EVENT_ATTRIBUTES_LIST = %i[availability_status availability_status_error].freeze
 
+  # TODO: IGNORE_RAISE_EVENT_ATTRIBUTES_LIST will be added later
   IGNORE_RAISE_EVENT_LIST = {
-    "Application"    => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST + %i[_superkey],
-    "Authentication" => IGNORE_RAISE_EVENT_ATTRIBUTES_LIST,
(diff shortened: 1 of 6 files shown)


lpichler:
and Is this PR returning back the issue from https://issues.redhat.com/browse/RHCLOUD-12860 ?

app/models/concerns/event_concern.rb
```diff
@@ -8,9 +8,9 @@ module EventConcern
   IGNORE_RAISE_EVENT_ATTRIBUTES_LIST = %i[availability_status availability_status_error].freeze
 
   IGNORE_RAISE_EVENT_LIST = {
```
slemrmartin, line 10: Pls add a TODO comment

slemrmartin (changes requested):

lindgrenj6:
@lpichler yes, turns out we should have checked with subswatch before pushing the change, they need the messages. 

slack thread here: https://ansible.slack.com/archives/CKJF47RS9/p1617902983110100

app/models/concerns/event_concern.rb
lindgrenj6: I'm not sure what the `TODO` would be? I don't see us re-enabling this since subswatch needs it.

slemrmartin: I'm wondering it could be something like 
```
# TODO: IGNORE_RAISE_EVENT_ATTRIBUTES_LIST will be added later
```

lindgrenj6: added

slemrmartin (approved):

PR #378, opened 2021-04-12
Availability check refactoring + authentications status reset

slemrmartin:
Any changes to kafka messages to `platform.sources.event-stream` are not a subject of this PR..

Sources availability-status - Update

---

**Issue** #317 
[RHCLOUD-13475](https://issues.redhat.com/browse/RHCLOUD-13475)

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -27,6 +27,10 @@ class Source < ApplicationRecord
     default || endpoints.build(:default => true, :tenant => tenant)
   end
 
+  def has_endpoint?
+    endpoints.any?
+  end
+
   def super_key?
     app_creation_workflow == SUPERKEY_WORKFLOW
   end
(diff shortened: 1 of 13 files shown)


lindgrenj6 (approved):
LGTM, this will be nice to add to a job once we have sidekiq.

app/models/source.rb
```diff
@@ -36,15 +36,49 @@ def super_key_credential
     authentications.detect { |a| a.authtype == source_type.superkey_authtype }
   end
 
-  def remove_availability_status(source = nil)
-    return if source == :Application && endpoints.any?
+  # resets availability status and runs new availability check
+  def reset_availability
+    super
```
lpichler, line 41: I am thinking  whether is good to use "super" for calling methods from mixin. 
For me it is confusing or maybe we need to at least raise error if that super method exists. 

in addition it looks that this method is doing two things but not as name declares.

when you call Source.first.reset_availability - it is doing reset_availability and availability_check
and when you call Source.first.reset_availability! -  it is doing reset_availability only (and save state)

maybe it will be better to use different  names and avoid to use super for calling method from mixin.

what do you think about it ?

slemrmartin: > I am thinking whether is good to use "super" for calling methods from mixin.
For me it is confusing or maybe we need to at least raise error if that super method exists.

it's a feature of the concern we can use super (but I didn't know about it too, so I understand your concerns), but the other option is to repeat the same code in all models.

> when you call Source.first.reset_availability - it is doing reset_availability and availability_check
and when you call Source.first.reset_availability! - it is doing reset_availability only (and save state)

the Source.reset_availability! is doing also both, it calls the availability check too (because the method reset_availability is overriden in the Source model). Although I understand what are you saying, my intention is to use one name for all models instead of the previous concept.

@lindgrenj6 your thoughts?

lpichler (approved):
👍

PR #379, opened 2021-04-13
Add Sidekiq for scheduled jobs

lindgrenj6:
This is just adding sidekiq, updated the clowdapp inline. 

I set up ActiveJob to use sidekiq as the backend as well - that way we can use standard rails generators as well as being able to change out the backend as needed. 

e2e-deploy changes for ci/qa are here: https://github.com/RedHatInsights/e2e-deploy/pull/2961

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,8 +4,8 @@ plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
 
 # Parser for Clowder config in ENV['ACG_CONFIG'] path
-gem 'clowder-common-ruby',  '~> 0.2.2'
 gem 'cloudwatchlogger',     '~> 0.2.1'
+gem 'clowder-common-ruby',  '~> 0.2.2'
 gem 'insights-api-common',  '~> 5.0', '>= 5.0.2'
 gem 'jbuilder',             '~> 2.0'
 gem 'json-schema',          '~> 2.8'
@@ -18,6 +18,7 @@ gem 'puma',                 '~> 4.3.5', '>= 4.3.5'
 gem 'pundit',               '~> 2.1'
 gem 'rack-cors',            '>= 1.1.1', '~> 1.1'
 gem 'rails',                '~> 5.2.2'
+gem 'sidekiq',              '~> 5.2.2'
 gem 'sprockets',            '~> 4.0'
 
 group :development, :test do
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,7 +1,17 @@
+redis_host = Sources::Api::ClowderConfig.instance["inMemoryDbHostname"]
+redis_port = Sources::Api::ClowderConfig.instance["inMemoryDbPort"]
+redis_password = Sources::Api::ClowderConfig.instance["inMemoryDbPassword"]
+
 Sidekiq.configure_server do |config|
-  config.redis = {:url => "redis://#{ENV["REDIS_HOST"] || "localhost"}:#{ENV["REDIS_PORT"] || 6379}/0"}
+  config.redis = {
+    :url      => "redis://#{redis_host || "localhost"}:#{redis_port || 6379}/0",
+    :password => redis_password
+  }
 end
 
 Sidekiq.configure_client do |config|
-  config.redis = {:url => "redis://#{ENV["REDIS_HOST"] || "localhost"}:#{ENV["REDIS_PORT"] || 6379}/0"}
+  config.redis = {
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -18,6 +18,7 @@ gem 'puma',                 '~> 4.3.5', '>= 4.3.5'
 gem 'pundit',               '~> 2.1'
 gem 'rack-cors',            '>= 1.1.1', '~> 1.1'
 gem 'rails',                '~> 5.2.2'
+gem 'sidekiq',              '~> 5'
```
slemrmartin, line 21: it uses version 6.2.1, at least minor version should be specified with `~>`

bin/bundle
```diff
@@ -1,3 +1,114 @@
 #!/usr/bin/env ruby
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
-load Gem.bin_path('bundler', 'bundle')
+# frozen_string_literal: true
```
slemrmartin, line 2: the whole change doesn't look like sidekiq specific code

deploy/clowdapp.yaml
```diff
@@ -19,6 +19,47 @@ objects:
   spec:
     envName: ${ENV_NAME}
     deployments:
+    - name: sidekiq
```
slemrmartin, line 22: I'm surprised it isn't a forked process inside the sources-api pod but needs separated deployment

deploy/clowdapp.yaml
```diff
@@ -19,6 +19,47 @@ objects:
...
+        - name: PATH_PREFIX
+          value: ${PATH_PREFIX}
+        - name: SECRET_KEY_BASE
+          valueFrom:
+            secretKeyRef:
+              name: sources-api-secrets
+              key: secret-key
+        resources:
+          limits:
+            cpu: ${CPU_LIMIT}
```
slemrmartin, line 58: If another pod is needed and sidekiq requires less memory, resources should have their own parameters

slemrmartin (changes requested):

slemrmartin (changes requested):
.

deploy/clowdapp.yaml
```diff
@@ -19,6 +19,47 @@ objects:
...
+        env:
+        - name: APP_NAME
+          value: ${APP_NAME}
+        - name: DB_POOL_SIZE
+          value: ${DB_POOL_SIZE}
+        - name: RAILS_LOG_LEVEL
+          value: ${LOG_LEVEL}
+          # TODO: It never can be blank!
+        - name: CLOUD_METER_AVAILABILITY_CHECK_URL
+          value: http://${CLOUD_METER_API_HOST}:${CLOUD_METER_SOURCES_API_PORT}${CLOUD_METER_SOURCES_API_AVAILABILITY_CHECK_PATH}
```
slemrmartin, line 40: please add CLOUD_METER_API_SCHEME like https://github.com/RedHatInsights/sources-api/pull/378/files#diff-50b41779d45e37968dc1bef161feebad3447f6c3cf72cfee62db774ecd5dfa4dR56

deploy/clowdapp.yaml
lindgrenj6: Yes it is a dedicated pod with a different entrypoint. Allows for scaling as well.

bin/bundle
lindgrenj6: It is not - but `bin/bundle` was not generated with bundler so in order to get `bin/sidekiq` generated it had to regenerate it.

Gemfile
lindgrenj6: I made it match our rails version - I will update it to `~> 5.2`, we can update it to 6 when we upgrade our rails version.

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,7 @@
+Sidekiq.configure_server do |config|
+  config.redis = {:url => "redis://#{ENV["REDIS_HOST"] || "localhost"}:#{ENV["REDIS_PORT"] || 6379}/0"}
```
slemrmartin, line 2: It has to use Clowder values from ACG_CONFIG env

lindgrenj6:
We really need to get the bot fixed so it doesn't endlessly spam when
PR's aren't mergable lol but when we get time for that I do not know.

-- 
Jacob Lindgren

Red Hat
***@***.***     M: 402-841-6173

slemrmartin (changes requested):
LGTM, only update the the sidekiq initializer

config/initializers/sidekiq.rb
slemrmartin: or this is the initializer, so Sources::Api::ClowderConfig can be used

slemrmartin (approved):
:+1:

PR #381, opened 2021-04-15
Application's availability check callback

slemrmartin:
This PR removes sending of availability check requests to kafka -> operation workers if the Source doesn't have an endpoint.

---

- [x] **depends on** https://github.com/RedHatInsights/e2e-deploy/pull/2956
- [x] **depends on** https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/17892

[RHCLOUD-12898](https://issues.redhat.com/browse/RHCLOUD-12898)

### DIFF
--- a/app/models/concerns/availability_status_concern.rb
+++ b/app/models/concerns/availability_status_concern.rb
@@ -35,16 +35,16 @@ module AvailabilityStatusConcern
     save!
   end
 
-  # sets availability status from caller's status
+  # sets availability status from dependent's status
   # now Application -> Source
   #
-  # @param caller [ActiveRecord::Base]
-  def set_availability!(caller)
-    self.availability_status = caller.availability_status
-    if respond_to?(:availability_status_error) && caller.respond_to?(:availability_status_error)
-      self.availability_status_error = caller.availability_status_error
+  # @param dependent [ActiveRecord::Base]
+  def set_availability!(dependent)
+    self.availability_status = dependent.availability_status
+    if respond_to?(:availability_status_error) && dependent.respond_to?(:availability_status_error)
+      self.availability_status_error = dependent.availability_status_error
     end
-    self.last_checked_at = caller.last_checked_at
+    self.last_checked_at = dependent.last_checked_at
     save!
   end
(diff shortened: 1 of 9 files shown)


app/models/source.rb
```diff
@@ -36,18 +36,23 @@ def super_key_credential
     authentications.detect { |a| a.authtype == source_type.superkey_authtype }
   end
 
+  # reset from Application can't invoke new check in Application
+  def reset_availability!(availability_check: true)
+    reset_availability(:availability_check => availability_check)
+  end
+
```
lindgrenj6, line 43: I'm not sure I see the reason for the `!` at the end of this method - if all it's doing is wrapping the non-! method why have it? Unless we think there is going to be more logic in this method?

lindgrenj6 (commented):
This looks good for the most part - just one little question on why we have the indirect reset_availability! method.

app/models/source.rb
slemrmartin: @lindgrenj6 this is overriden method `reset_availability!` from the concern, which calls `save!`, but only for Source model it's not wanted

lindgrenj6 (approved):

app/models/concerns/availability_status_concern.rb
```diff
@@ -33,4 +33,17 @@ def reset_availability!
     reset_availability
     save!
   end
+
+  # sets availability status from caller's status
+  # now Application -> Source
+  #
+  # @param caller [ActiveRecord::Base]
+  def set_availability!(caller)
```
lpichler, line 41: https://apidock.com/ruby/v2_6_3/Kernel/caller is also method, I am thinking whether it is not confusing.

slemrmartin: maybe, do you have some suggestion about other name?

lpichler: It looks that `caller` is object or  resource so maybe it can be `caller_resource` or `caller_object` or just `resource`.

lpichler (approved):

PR #383, opened 2021-04-15
Add Pause support for Applications

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-9850

EDIT: 4/20/21 - this is ready to go. there is one really odd thing going on with the specs - its something to do with how AR relations form - it won't work unless I manually assign the auth to the application for the first spec. If there is a way to get around that I'm all ears!

This PR:
- [x] adds the `discard` gem
- [x] adds a `Pausable` concern which includes the module + sets the column to `paused_at`
- [x] adds `POST /applications/:id/pause` and `POST /applications/:id/unpause` endpoints for handling the pausing
- [x] updates the OpenAPI schema with the changes. 
- [x] adds before/after_discard callbacks to handle source discard status
- [x] discards dependent authentications
- [x] posts pause/unpause events via a job
- [x] specs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 # Parser for Clowder config in ENV['ACG_CONFIG'] path
 gem 'cloudwatchlogger',     '~> 0.2.1'
 gem 'clowder-common-ruby',  '~> 0.2.2'
+gem 'discard',              '~> 1.2'
 gem 'insights-api-common',  '~> 5.0', '>= 5.0.2'
 gem 'jbuilder',             '~> 2.0'
 gem 'json-schema',          '~> 2.8'
--- a/app/controllers/api/v3x1.rb
+++ b/app/controllers/api/v3x1.rb
@@ -6,7 +6,6 @@ module Api
       end
     end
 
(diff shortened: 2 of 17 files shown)


slemrmartin (changes requested):
Pausing shouldn't influence the availability checking (discussed on Slack)
Also it doesn't change the connection data of Source or associations
In that case I think the `paused_at` column has to be added to the AvailabilityStatusConcern's `IGNORE_LIST` thus it doesn't reset the av. status

lindgrenj6:
@slemrmartin 
>Pausing shouldn't influence the availability checking (discussed on Slack)

It shouldn't, this does not change the API responses at all, so the availability checkers should run for even paused sources.

> Also it doesn't change the connection data of Source or associations

Not sure what you mean here - if you mean it doesn't pause/unpause source+assocations it does, I have `(before|after)_discard` callbacks that discard/undiscard the subresources.

>In that case I think the paused_at column has to be added to the AvailabilityStatusConcern's IGNORE_LIST thus it won't reset the av. status

Yes I need to do that still.

slemrmartin (approved):
:+1: LGTM (the openapi comment is only question, not change request)

lpichler (approved):

lpichler (approved):

PR #385, opened 2021-04-20
add in_progress to availability_status enum for Application model

lindgrenj6:
Forgot to allow `Application` to be allowed "in progress", this is necessary for Cost to post in the status to us. 

cc @syncrou @dccurtis


slemrmartin (approved):
nice, I was curious if it was intention...so it wasn't :)

PR #390, opened 2021-05-07
Internal API v2 - pagination

slemrmartin:
GET /internal/v1.0/tenants is not paginated

In production, there is >13000 records which causes >2s request duration

---

[RHCLOUD-11708](https://issues.redhat.com/browse/RHCLOUD-11708)

### DIFF
--- a/app/models/tenant.rb
+++ b/app/models/tenant.rb
@@ -6,5 +6,16 @@ class Tenant < ApplicationRecord
   def self.tenancy_enabled?
     ENV["BYPASS_TENANCY"].blank?
   end
-end
 
+  # as_json serialization for internal API's tenants_controller
+  # overrides Insights::API::Common::OpenApi::Serializer
+  def _schema(arg)
+    version = api_version_from_prefix(arg[:prefixes].first)
+    if version.to_f < 2.0
+      super
+    else
+      presentation_name = self.class.try(:presentation_name) || self.class.name
+      ::Sources::Api::InternalDocs.instance[version].definitions[presentation_name]
+    end
+  end
+end
--- a/private/doc/openapi-3-v2.0.json
+++ b/private/doc/openapi-3-v2.0.json
@@ -1,0 +1,209 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "description": "Sources Internal",
+    "version": "2.0.0",
+    "title": "Sources Internal",
+    "contact": {
+      "email": "support@redhat.com"
+    },
+    "license": {
+      "name": "Apache 2.0",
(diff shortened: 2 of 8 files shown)


private/doc/openapi-3-v2.0.json
```diff
@@ -0,0 +1,209 @@
...
+  },
+  "security": [
+    {
+      "UserSecurity": [
+
+      ]
+    }
+  ],
+  "paths": {
+  },
```
lindgrenj6, line 23: It really looks like the generator isn't working with this, the `paths` object should contain all of the `/internal` endopints - so lets just create one manually rather than modifying the generator.

app/models/tenant.rb
```diff
@@ -6,5 +6,16 @@ class Tenant < ApplicationRecord
   def self.tenancy_enabled?
     ENV["BYPASS_TENANCY"].blank?
   end
-end
 
+  # as_json serialization for internal API's tenants_controller
+  # overrides Insights::API::Common::OpenApi::Serializer
+  def _schema(arg)
```
lindgrenj6, line 12: Is this only needed for the generator? If so - lets remove it since the generator doesn't work for this and we're going to not use the generator anymore.

private/doc/openapi-3-v2.0.json
lindgrenj6: For the most part the rest of the schema looks fine - so maybe just manually add the paths to this?

lindgrenj6 (changes requested):
This is kind of exposing the problem with the generator - it didn't behave right with the `/internal` endpoints; so I think since we're going away from the generator for the main openapi schema we should just take what the generator gave us here and roll it by hand. 

It would probably be better to do that than spend time fixing the generator since we aren't going to use it anymore. I recommend this editor: https://editor.swagger.io/ it has tab completion and shortcuts to add objects/endpoints etc.

private/doc/openapi-3-v2.0.json
syncrou: I agree here.  If we can't trust the generator to do the right thing, let's just do a manual generation of the schema.

syncrou (commented):
This looks great.  I'm thinking it's time to retire the generator. It seems either way we'll need to fix the openapi.json spec.

private/doc/openapi-3-v2.0.json
slemrmartin: It was created manually so I have to add the /paths (at least it brokes the specs)

lindgrenj6 (approved):

PR #394, opened 2021-05-12
Internal SourcesController

slemrmartin:
Primary usage of the internal/v2.0/sources is by SourcesMonitor cronjobs to avoid tenant based multiple queries.

Btw (not related to this PR) I found either insights-api-common or acts-as-tenant are introducing N+1 queries problem when requesting Source.all (in both this and public SourcesController)

- [x] **depends on** #390 

---

[RHCLOUD-14101](https://issues.redhat.com/browse/RHCLOUD-14101)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -139,6 +139,7 @@ Rails.application.routes.draw do
 
     namespace :v2x0, :path => "v2.0" do
       resources :authentications, :only => [:show]
+      resources :sources,         :only => [:index]
       resources :tenants,         :only => [:index, :show]
     end
   end
--- a/app/controllers/internal/v2x0/mixins/index_mixin.rb
+++ b/app/controllers/internal/v2x0/mixins/index_mixin.rb
@@ -1,0 +1,37 @@
+module Internal
+  module V2x0
+    module Mixins
+      module IndexMixin
+        def index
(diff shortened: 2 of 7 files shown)


lindgrenj6:
Just need a rebase here - looks good from just looking at the last 2 commits since this is based on #390 . Tried to do it myself but not having luck.

lindgrenj6 (approved):

config/routes.rb
```diff
@@ -139,6 +139,7 @@ def put(*_args); end
 
     namespace :v2x0, :path => "v2.0" do
       resources :authentications, :only => [:show]
+      resources :sources,         :only => [:index]
       resources :tenants,         :only => [:index, :show]
```
lindgrenj6, line 143: apparently my comment didn't come with the review...great. 

Just wanted to validate that it looks like with https://github.com/RedHatInsights/sources-monitor/pull/30 it's not hitting the `/tenants` endpoint anymore - so it's not really needed. I'm ok with leaving it here though since we could potentially use it elsewhere, I just want to make sure I'm reading both of these PRs right.

slemrmartin: Yes, `/tenant` is not needed after the sources-monitor's PR will be applied

PR #398, opened 2021-05-18
Add Sidekiq job for availability message processing

lindgrenj6:
This is part of my hackathon work!

Basically moving the availability status listener's guts into a sidekiq job - so we can process more of them than one at a time. The biggest problem before was that we processed the update messages serially - so it could take quite a while to process if a bunch of them came back at once. This way we can increase the number of sidekiq workers and process them as fast as we want. 

cc @syncrou

### DIFF
--- a/app/jobs/availability_status_update_job.rb
+++ b/app/jobs/availability_status_update_job.rb
@@ -1,0 +1,38 @@
+class AvailabilityStatusUpdateJob < ApplicationJob
+  queue_as :availability
+
+  def perform(event, headers)
+    payload = JSON.parse(event)
+    model_class = payload["resource_type"].classify.constantize
+
+    validate_resource_type(model_class)
+
+    record_id = payload["resource_id"]
+    object = model_class.find(record_id)
+
+    options = {
+      :availability_status => payload["status"],
+      :last_checked_at     => Time.now.utc
+    }
+
+    options[:availability_status_error] = payload["error"] if %(Endpoint Application).include?(model_class.name)
+    options[:last_available_at] = options[:last_checked_at] if options[:availability_status] == 'available'
+
+    object.update!(options)
+
+    object.raise_event_for_update(options.keys, headers)
+  rescue NameError
+    Sidekiq.logger.error("Invalid resource_type #{payload["resource_type"]}")
+  rescue ActiveRecord::RecordNotFound
+    Sidekiq.logger.error("Could not find #{model_class} with id #{record_id}")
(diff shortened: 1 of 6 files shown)


app/jobs/availability_status_update_job.rb
```diff
@@ -0,0 +1,38 @@
...
+    Sidekiq.logger.error("Could not find #{model_class} with id #{record_id}")
+  rescue ActiveRecord::RecordInvalid
+    Sidekiq.logger.error("Invalid status #{payload["status"]}")
+  rescue => e
+    Sidekiq.logger.error(["Something is wrong when processing Kafka message: ", e.message, *e.backtrace].join($RS))
+  end
+
+  def validate_resource_type(model_class)
+    # For security reason only accept explicitly listed models
+    raise NameError unless [Application, Endpoint, Source].include?(model_class)
```
slemrmartin, line 36: although the column is not actively used, Authentication also supports availability check so it should be included in the list

app/jobs/availability_status_update_job.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    record_id = payload["resource_id"]
+    object = model_class.find(record_id)
+
+    options = {
+      :availability_status => payload["status"],
+      :last_checked_at     => Time.now.utc
+    }
+
+    options[:availability_status_error] = payload["error"] if %(Endpoint Application).include?(model_class.name)
```
slemrmartin, line 18: also Authentication

spec/jobs/availability_status_update_job_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+  subject { AvailabilityStatusUpdateJob }
+
+  context "when body contains valid resource_type and id" do
+    let(:endpoint) do
+      create(:endpoint, :role => "first", :default => true)
+    end
+    let(:resource_type) { "endpoint" }
+    let(:resource_id)   { endpoint.id.to_s }
+
+    before { allow(Time).to receive(:now).and_return(now) }
```
slemrmartin, line 19: using Timecop will look better

slemrmartin (commented):
looks good, just some small things

spec/jobs/availability_status_update_job_spec.rb
lindgrenj6: First time using it, but fixed it to use Timecop.

slemrmartin (approved):

PR #399, opened 2021-05-18
Add sidekiq job for requesting endpoint availability

lindgrenj6:
Moar hackathon work...

the `POST /sources/:id/check_availability` operation can slow to an absolute crawl during the sources-monitor runs mostly due to the fact that it's producing a message for every source, which can balloon the average request time to 1-2 seconds.

By queuing them in sidekiq we can process them in parallel and most importantly *out of the request* so the request time is still wicked fast. 

cc @syncrou 

EDIT: I imported the CI db locally and ran a test to see how much faster this will be - I ran both checkers like so: 

```
time bash -c "bundle exec bin/availability_checker available && bundle exec bin/availability_checker unavailable" >/dev/null
```

the results are as such:
old code (produce message in the request): 
```
real    0m22.084s
user    0m1.631s 
sys     0m0.301s 
```

new code (with sidekiq taking a bit longer to process, but it produces 5 messages at once but not in the request):
```
real    0m5.298s 
user    0m1.577s 
sys     0m0.330s 
```

This will improve more once the tenancy prs are merged too! Should be great!

### DIFF
--- a/app/jobs/kafka_publish_job.rb
+++ b/app/jobs/kafka_publish_job.rb
@@ -1,0 +1,13 @@
+class KafkaPublishJob < ApplicationJob
+  queue_as :default
+
+  def perform(topic, event, payload)
+    Sidekiq.logger.info("Publishing #{event} to #{topic}...")
+
+    Sources::Api::Messaging.client.publish_topic(
+      :service => topic,
+      :event   => event,
+      :payload => payload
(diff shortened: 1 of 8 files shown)


slemrmartin (approved):
:+1: nice, it means publishing to kafka is very inefective, interesting

PR #403, opened 2021-05-24
Retry connecting to the db if the listener runs into a connection error

lindgrenj6:
So I just noticed this on prod - the listener apparently lost connection to the db and since it's just running in a thread it apparently doesn't manually reconnect. 

The little handler there tries to re-establish the connection before failing _again_ and being caught by the final rescue clause again. This way if it's just a blip it'll reconnect and retry.

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -59,6 +59,11 @@ class AvailabilityStatusListener
     Rails.logger.error("Could not find #{model_class} with id #{record_id}")
   rescue ActiveRecord::RecordInvalid
     Rails.logger.error("Invalid status #{payload["status"]}")
+  rescue ActiveRecord::StatementInvalid, PG::UnableToSend
+    Rails.logger.error("Ran into error connecting to db - retrying.")
+    ActiveRecord::Base.establish_connection
+
+    retry
   rescue => e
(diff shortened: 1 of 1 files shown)


lib/availability_status_listener.rb
```diff
@@ -59,6 +59,11 @@ def process_event(event)
     Rails.logger.error("Could not find #{model_class} with id #{record_id}")
   rescue ActiveRecord::RecordInvalid
     Rails.logger.error("Invalid status #{payload["status"]}")
+  rescue ActiveRecord::StatementInvalid, PG::UnableToSend
+    Rails.logger.error("Ran into error connecting to db - retrying.")
+    ActiveRecord::Base.establish_connection
+
+    retry
```
lpichler, line 66: I am just thinking whether we need count and set limit how many times we can retry but maybe we can try it without limit.

lindgrenj6:
The good news is that if the `establish_connection` call fails (like the db
is unreachable still) - it will throw a different exception that gets
caught by the blanket rescue case at the bottom. So it only tries to
re-establish the connection one time and if it fails it skips the message.

That _seems_ to be what we would want - if there is a blip in connectivity
we skip for a while, then once it comes back it'll continue as normal.
--
Jacob Lindgren

Red Hat
***@***.***     M: 402-841-6173

slemrmartin (approved):

PR #406, opened 2021-05-26
Add deployment to ephemeral env and run smoke tests on pr_check

zsadeh:


slemrmartin (commented):
@zsadeh this code returns error and doesn't generate the xml

the sources-api pod is not ready. I think custom bonfire config has to be provided, not sure what exactly.

Anyway when I tried to run rspec tests through pr_check, it ended with out of memory error.

slemrmartin (approved):
great! it looks good now

zsadeh:
@slemrmartin This is fixed now (Brandon Squizatto found issue in bonfire)

PR #407, opened 2021-05-26
Validate incoming `platform.sources.status` messages by requiring x-rh-identity header

lindgrenj6:
So troubleshooting issues with cost + swatch I found that cloudigrade was not sending the x-rh-identity header (here: https://github.com/cloudigrade/cloudigrade/blob/13a91f0f7aaf811ba38e5ca54c6e005c19694796/cloudigrade/util/redhatcloud/sources.py#L256) so that was causing cost to not be happy.

This PR basically adds a check on that header - if it isn't there we'll log a NASTY error message and skip producing the message. 

This is probably the ideal solution - the client apps (e.g. swatch, catalog, us) will think the messages are missing, but we are going to have log messages saying they're missing a header. Open to discussion otherwise.

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -5,6 +5,7 @@ class AvailabilityStatusListener
   SERVICE_NAME = "platform.sources.status".freeze
   GROUP_REF = "sources-api-status-worker".freeze
   EVENT_AVAILABILITY_STATUS = "availability_status".freeze
+  REQUIRED_HEADERS = %w(x-rh-identity).freeze
 
   attr_accessor :messaging_client_options, :client
 
@@ -38,6 +39,11 @@ class AvailabilityStatusListener
   def process_event(event)
     Rails.logger.info("Kafka message #{event.message} received with payload: #{event.payload}")
     return unless event.message == EVENT_AVAILABILITY_STATUS
+
+    if (REQUIRED_HEADERS - event.headers.keys).any?
+      Rails.logger.error("Kafka message #{event.message} missing required header(s) [#{REQUIRED_HEADERS.join(",")}], found: [#{event.headers.keys.join(',')}]; returning.")
+      return
+    end
 
     payload = JSON.parse(event.payload)
     model_class = payload["resource_type"].classify.constantize
--- a/spec/lib/availability_status_listener_spec.rb
+++ b/spec/lib/availability_status_listener_spec.rb
@@ -5,11 +5,10 @@ RSpec.describe AvailabilityStatusListener do
   let(:status)     { "unavailable" }
   let(:reason)     { "host unreachable" }
   let(:now)        { Time.new(2020).utc }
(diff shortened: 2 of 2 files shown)


lib/availability_status_listener.rb
```diff
@@ -39,6 +40,11 @@ def process_event(event)
     Rails.logger.info("Kafka message #{event.message} received with payload: #{event.payload}")
     return unless event.message == EVENT_AVAILABILITY_STATUS
 
+    if (REQUIRED_HEADERS - event.headers.keys).any?
+      Rails.logger.error("Kafka message missing required header(s): #{(REQUIRED_HEADERS - event.headers.keys).join(',')}; returning.")
```
slemrmartin, line 44: I would add some identification, i.e. `event.message` (and maybe the payload too?)

slemrmartin (changes requested):
Looks good to me, just provide more info to log for faster identification

lib/availability_status_listener.rb
lindgrenj6: That gets logged on line 40 right away at the beginning of processing. Is that ok?

slemrmartin: When the event.message will be there too, then you can filter messages by the string in the Kibana so it will be more comfortable for troubleshooting

lindgrenj6: Ahh yeah good point, added!

lib/availability_status_listener.rb
```diff
@@ -39,6 +40,11 @@ def process_event(event)
     Rails.logger.info("Kafka message #{event.message} received with payload: #{event.payload}")
     return unless event.message == EVENT_AVAILABILITY_STATUS
 
+    if (REQUIRED_HEADERS - event.headers.keys).any?
+      Rails.logger.error("Kafka message #{event.message} missing required header(s): #{(REQUIRED_HEADERS - event.headers.keys).join(',')}; returning.")
```
syncrou, line 44: Should we say:  ".. missing required header(s) #{REQUIRED_HEADERS}: found: #{event.headers.keys.join(',')}; returning"

So the `REQUIRED_HEADERS` are included in the log message?

slemrmartin (approved):

slemrmartin:
@lindgrenj6 PR check error: 
```
When PR is not rebased, Jenkins will create a 'merge commit' locally and this ref does not exist in the repo
```

lib/availability_status_listener.rb
lindgrenj6: I like that idea - that way we can see the differences in formats they would have without having to debug in.

lindgrenj6:
@slemrmartin according to the devprod team there is a bug in clowder right now - pr_checks are failing. maybe they'll be fixed here soon. Probably safe to merge before then though.

lpichler (approved):

PR #409, opened 2021-06-01
Add optionalDependencies rbac to clowdapp.yaml

zsadeh:


slemrmartin (changes requested):
@zsadeh please fix the indentation (there shouldn't be 2 spaces before the dash)

slemrmartin (approved):

PR #410, opened 2021-06-02
Add PSK support

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-12955

\# TODO:
- [x] specs
- [x] validate message producing works when authing with PSK, forge x-rh-identity if necessary!

**EDIT:** So, the biggest part about this is that we probably need to have backward compatibility with the x-rh-id header + the new PSK related headers. 

To get around this I overrode the `Insights::API::Common::Request.current_forwardable` class method to add any new fields we want to - currently just forwarding the account number since they will have the PSK mounted as a secret in their namespace. 

----

I am probably going to do the kafka availability status listener changes (no more required x-rh-id header, required account number) in a separate PR that way this one doesn't get huge. 

So this PR basically just adds internal communication via PSK and passes on the new headers when they are used.

### DIFF
--- a/app/policies/default_policy.rb
+++ b/app/policies/default_policy.rb
@@ -1,8 +1,9 @@
 class DefaultPolicy
-  attr_reader :user, :record
+  attr_reader :request, :key, :record
 
-  def initialize(user, record)
-    @user = user
+  def initialize(context, record)
+    @request = context.request
+    @key = context.key
     @record = record
   end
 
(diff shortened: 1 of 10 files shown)


app/policies/default_policy.rb
```diff
@@ -32,16 +33,30 @@ def destroy?
...
-    user.system.present? || user.user&.org_admin? || write_access?
+    # TODO: remove org_admin after everyone has moved over.
+    # Maybe even remove the `system` check
+    psk_matches? || request.system.present? || request.user&.org_admin? || write_access?
+  end
+
+  def psk_matches?
+    return false if self.class.pre_shared_key.nil?
+
+    self.class.pre_shared_key == key
```
slemrmartin, line 44: In case of receptor, they have array of keys as their env variable (in the vault) and separate key-value in the Vault for each team. What do you think about it?

slemrmartin: stage-receptor-stage.yml: 
```
- provider: vault-secret
  path: insights/secrets/insights-stage/receptor-stage/receptor-psks
  version: 4
  annotations:
    qontract.recycle: "true"
```

stage-sources-stage.yml:
```
- provider: vault-secret
  path: insights/secrets/insights-stage/receptor-stage/sources-psk
  version: 1
  annotations:
    qontract.recycle: "true"

```

lindgrenj6: Ahhh interesting - I didn't know that, I'll update it to take an array. That way we can follow that implementation.

slemrmartin (approved):

lpichler (approved):

lpichler:
LGTM, can it be merged regardless failure in ci.ext.devshift.net PR build?

lindgrenj6:
@lpichler yeah, the eph cluster appears to be having issues, looking at the error it's not an "us" problem its a clowder problem:

```
15:38:02 Error from server (Forbidden): clowdenvironments.cloud.redhat.com is forbidden: User "system:serviceaccount:ephemeral-base:ephemeral-bot" cannot list resource "clowdenvironments" in API group "cloud.redhat.com" at the cluster scope
```

PR #412, opened 2021-06-04
Add support and backwards compatibility for x-rh-sources-psk on the availability status listener

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-14523

As we add the `x-rh-sources-psk` header - we need to maintain backwards compatibility on the `platform.sources.status` queue event processor. 

This was a little tough to implement - since basically we need to allow `(x-rh-id | x-rh-sources-account-number)`. To maintain backwards compatibility - if the user passes in the account number but *NOT* the x-rh-id, we generate one and pass it on.

### DIFF
--- a/lib/availability_status_listener.rb
+++ b/lib/availability_status_listener.rb
@@ -5,7 +5,6 @@ class AvailabilityStatusListener
   SERVICE_NAME = "platform.sources.status".freeze
   GROUP_REF = "sources-api-status-worker".freeze
   EVENT_AVAILABILITY_STATUS = "availability_status".freeze
-  REQUIRED_HEADERS = %w(x-rh-identity).freeze
 
   attr_accessor :messaging_client_options, :client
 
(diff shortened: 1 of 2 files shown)


lib/availability_status_listener.rb
```diff
@@ -40,15 +39,42 @@ def process_event(event)
...
     AvailabilityStatusUpdateJob.perform_later(event.payload, event.headers)
   end
 
+  # hash of "groups" where each value is an array of "one of" headers.
+  # each group needs to have at least one header from each group
+  REQUIRED_HEADER_GROUPS = {
+    :identity => %w[x-rh-identity x-rh-sources-account-number]
+  }.freeze
+
+  def get_missing_headers(headers)
```
lpichler, line 71: probably `get_missing_headers` -> `missing_headers`  is enough

lpichler (approved):

slemrmartin (commented):
I'm not sure if I understand the concept of this PR. If someone implements the `x-rh-sources-account-number`, it means it also implemented the psk, or not? Why will someone switch just to the account-number header without the psk header?

lindgrenj6:
Because based on some of the consumers (cost/swatch) their listeners break
if we don't have x-rh-identity present.

This is more of a backward-compatible shim for now, until we can confirm
that everyone doesn't need the x-rh-identity header anymore, or implements
workarounds so that the consumers don't break without it.

--
Jacob Lindgren

Red Hat
***@***.***     M: 402-841-6173

slemrmartin (approved):

PR #418, opened 2021-06-23
Add support for SOURCES_PSKS secret if it exists in the namespace

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-14651

Clowdapp change for ^, the `optional: true` denotion basically says "still spin up the deployment even if the secret isn't there, otherwise we would end up with a pod initialization error.


syncrou (approved):
LGTM 👍

syncrou:
@lpichler Can you take a look at this? Since it affects Prod I'd like more 👀 on it.

PR #419, opened 2021-06-28
pluralize SOURCES_PSKS env var read in DefaultPlicy to match ENV var

lindgrenj6:
Me being the derp that I am edited it to plural `PSKS` everywhere except the actual code.

PR #420, opened 2021-06-28
Change where the SOURCES_PSKS env var gets populated from, storing it…

lindgrenj6:
… in the `sources-api-secrets` secret in vault.

Would like to get #419 in too - that way I can get this working on stage.


syncrou (commented):
Is this name a requirement? Just asking to understand why the old name didn't work

syncrou (approved):
Ok. I understand this note

PR #422, opened 2021-06-29
Add default PSK value for ephemeral environment + local environment

lindgrenj6:
~Just using a pre-canned UUID for now - going to potentially change it to a non-generated one if I can follow up how to.~

Found out about the ability to generate strings with bonfire, nice. Just made it a randomly generated 16 character key.


lpichler (approved):
Nice!

PR #431, opened 2021-08-18
Generate `x-rh-sources-account-number` header when not present

lindgrenj6:
A few things found while doing integration work with swatch + cost:
1. The fact that the PSK changes _on every deploy_ is kind of a PITA in reality, so I'm changing the resource to always use the same string: `thisMustBeEphemeralOrMinikube`. Open to a funnier name if you have one!
2. Cost management still needs the x-rh-identity for some of the other fields in it (who would have thought that not only the account number would be used right?) so I changed the availability status listener to not overwrite it if present.
3. And the most important oversight I made during development was that I forgot that yes, sometime stuff can come from the outside still and those requests *will not* have the `x-rh-sources-account-number` header, so we need to reach into the `x-rh-identity` and pull out the account number so we can populate the internal header. This will only happen if the sources header isn't there but the x-rh-id is.

cc @maskarb @abellotti changes here for the account-number header generation on reqs/messages _without_ the x-rh-sources-account-number header.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -12,7 +12,7 @@ objects:
   stringData:
     encryption-key: "${ENCRYPTION_KEY}"
     secret-key: "${SECRET_KEY}"
-    psks: "${PSK}"
+    psks: "thisMustBeEphemeralOrMinikube"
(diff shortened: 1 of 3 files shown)


lpichler (approved):
LGTM 👍

PR #432, opened 2021-08-18
add docker-compose file

maskarb:
This PR adds a docker-compose file which runs kafka, zookeeper, sources-db, and sources-server for use by teams that need to test their integrations with sources-api locally. Using docker-compose gives us a simpler local setup for integration testing.

This file enables a user to spin up sources (and connect to a specific docker network, if they choose) like:
```
export DOCKER_NETWORK_NAME={network-name} (optional)
docker-compose up -d
```

This PR did require a change to `docker-assets/seed_database` in order for the sources-server to spin up correctly in docker. I know this file is used in the clowdapp, so we need to ensure the changes made do not affect stage/prod in any way.


docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - KAFKA_ZOOKEEPER_CONNECT=zookeeper:32181
+      - KAFKA_AUTO_CREATE_TOPICS_ENABLE=true
+
+  sources-server:
+    container_name: sources_server
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: ["run_rails_server"]
+    entrypoint: ["entrypoint", "seed_database"]
```
lindgrenj6, line 46: Since we're overriding the `CMD` in the Dockerfile anyway (compose's `entrypoint` does it implicitly) - can we just remove it here and make the entrypoint:
```suggestion
    entrypoint: ["entrypoint", "seed_database", "run_rails_server"]
```

test.json
```diff
@@ -0,0 +1,30 @@
+{
```
lindgrenj6, line 1: Can we rename this to `.docker-compose-config.json`? That way it doesn't show up in ls output.

~Also would like to add this to the `.dockerignore` file~ NVM that wouldn't work since it's needed for compose. It's probably fine.

docker-assets/seed_database
```diff
@@ -9,6 +9,25 @@ function write_encryption_key() {
...
+}
+
 write_encryption_key
 
+# Wait for postgres to be ready
+check_svc_status $DATABASE_HOST $DATABASE_PORT
+
 bundle exec rake db:migrate db:seed
+
+exec "$@"
```
lindgrenj6, line 33: So I'm pretty sure I get why this is needed - its basically so we can run `entrypoint seed_database run_rails_server` rather than `entrypoint seed_database && entrypoint run_rails_server`. I just want to validate that it doesn't exit with exit code 1 without any args which would indicate failure in our initContainer. I'll test it locally and validate.

lindgrenj6 (commented):
Thanks for the PR! Just a few cosmetic things and one clarification I need to make. But other than that it looks good.

test.json
maskarb: Done!

docker-assets/seed_database
maskarb: Yep! Without this, we'd exit locally when the seed would run.

lindgrenj6 (approved):
I validated this works in my local ENV! Waiting for the sidekiq commit and I'll merge it.

lindgrenj6:
Merging this even though theres a failure - the eph env spins up sources fine but is then failing on rbac. I think its fine on our end.

PR #442, opened 2021-09-01
Remove org_admin from generated header when psk is used

lindgrenj6:
Per slack apparently cost's integration doesn't like the generated xrhid. Removing the org_admin line from it. Will merge when I get confirmation.


lindgrenj6:
Closing this since this wasn't the issue.

PR #443, opened 2021-09-15
Add a proxy so we can send requests to the go svc selectively

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-16123

\# TODO:
- [x] implement proxy with faraday
- [x] support GET/PATCH/POST/DELETE
- [x] add configmap to clowdapp template

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@
 
 # Local Bonfire (Clowder deploy) settings
 /config.yaml
+/proxy_rules/rules.json
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 gem 'cloudwatchlogger',     '~> 0.2.1'
 gem 'clowder-common-ruby',  '~> 0.2.3'
 gem 'discard',              '~> 1.2'
+gem 'faraday',              '~> 1.7.2'
 gem 'insights-api-common',  '~> 5.0', '>= 5.0.6'
 gem 'jbuilder',             '~> 2.0'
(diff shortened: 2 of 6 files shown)


lpichler:
@lindgrenj6 it works great just one thing - it looks it doesn't forward parameters like `filter`:

for example
```
GET http://localhost:3002/api/sources/v3.1/application_types?filter[display_name]=Remediations
```

most likely filtering doesn't work as it was described [here](https://github.com/RedHatInsights/sources-api-go/pull/14#issuecomment-930406943) (but for filter parameter) but despite it I tried to inspect `filters` [here](https://github.com/RedHatInsights/sources-api-go/blob/main/dao/application_type_dao.go#L39) variable in debugger and it was empty.

lindgrenj6:
@lpichler Changed `request.path` -> `CGI.unescape(request.fullpath)`. 

```
[----] I, [2021-10-01T12:15:03.021882 #4045408:2ca3918]  INFO -- : Proxying request [/api/sources/v3.1/application_types?filter[display_name]=Remediations] to go svc at [http://localhost:8000]
```

It does look like filtering just isn't working anymore on the Go side anymore (if it ever did?) though. Even when it is getting set in the `applyFilters` function it isn't generating the `WHERE` in the SQL statement. We should probably change the way that function works since it appears to just straight up not work now.

lindgrenj6:
Fixed sorting/filtering in this commit over there: https://github.com/RedHatInsights/sources-api-go/pull/14/commits/bbed6ac5b386f262f3cb440e7c51eaf21ed225db

lpichler (approved):
thanks, Looks great 👍

PR #448, opened 2021-09-24
Lock down `system` header authorization

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-16318

\# TODO:
- [x] lock down `system` to only POST requests (per requirements)
- [x] ~only allow reads of that specific CN's objects instead of everything in the tenant~ doing something like this in a follow-up
- [x] fix tests to work without org_admin :dizzy_face:


lpichler (approved):

app/policies/mixins/write_policy_mixin.rb
```diff
@@ -1,9 +1,12 @@
 module WritePolicyMixin
   def create?
+    admin? || system?
```
lpichler, line 3: if I understand well, `admin?` can do API request to RBAC service - so in order to avoid some requests - is it beneficial to do `system?` first ? 

(Not sure whether order is intended)

lindgrenj6: So I couldn't remember so I went back and tried again - turns out there was a spec failure and that failure was the fact that it blew up when the x-rh-identity wasn't there. I think I flipped the order initially due to the fact I hadn't disabled RBAC globally yet.

I added a rescue on the `IdentityError` exception that gets thrown and returned false - since it obviously isn't system-authorized if there isn't a x-rh-identity header.

lpichler:
@lindgrenj6  I forgot to ask:

Do we need to remove `DISABLE_ORG_ADMIN` from  `deploy/clowdapp.yaml` and `docker-compose.yml` ?

lindgrenj6:
@lpichler yeah we probably should eventually - but no rush there since its just busy work. I'll take care of it once we get the proxy + the go svc deployed and running.

PR #450, opened 2021-09-29
[wip] Lock down cert-auth's viewing to only source/endpionts + the satellite source type

lindgrenj6:
Continuing work done in #448 here

\# TODO:
- [x] disallow index/show on applications/authentications if using cn authentication
- [ ] limit index/show on source/endpoint to source_type satellite


lpichler (approved):
LGTM 👍

PR #460, opened 2021-11-11
Changing Azure "SuperKey" to use a button rather than Superkey Worker

lindgrenj6:
Reverted the old work I did and just added the new authtype for the wizard + the URLs to serve for the button.

\# TODO: 
- [x] waiting on a public URL for stage. WIll mark as ready when that is there and this PR is updated.


lindgrenj6:
Merging to deploy app meta data for CM azure manual mode wizard.

PR #461, opened 2021-11-16
Add IBM Source Type

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-16490

@rvsia the icon_url would need to be added on your side right?


rvsia:
@lindgrenj6 you can go ahead and add it to here: https://github.com/RedHatInsights/frontend-assets just follow the same pattern as for the other icons

rvsia (approved):

PR #466, opened 2021-12-14
[RHCLOUD-17299] Add rh-marketplace application and source types

MikelAlejoBR:
Attempts to add Red Hat Marketplace as an application type and a source type.

[[RHCLOUD-17299]](https://issues.redhat.com/browse/RHCLOUD-17299)

### DIFF
--- a/db/seeds/application_types.yml
+++ b/db/seeds/application_types.yml
@@ -24,14 +24,6 @@
     - project_id_service_account_json
     openshift:
     - token
-"/insights/platform/rh-marketplace":
-  :display_name: Red Hat Marketplace
-  :dependent_applications: []
-  :supported_source_types:
-  - rh-marketplace
-  :supported_authentication_types:
-    rh-marketplace:
-      - token
 "/insights/platform/cloud-meter":
   :display_name: RHEL management
   :dependent_applications: []
--- a/db/seeds/application_types.yml
+++ b/db/seeds/application_types.yml
@@ -24,6 +24,14 @@
     - project_id_service_account_json
     openshift:
     - token
+"/insights/platform/rh-marketplace":
+  :display_name: Red Hat Marketplace
+  :dependent_applications: []
+  :supported_source_types:
+  - rh-marketplace
+  :supported_authentication_types:
+    rh-marketplace:
+      - token
 "/insights/platform/cloud-meter":
   :display_name: RHEL management
   :dependent_applications: []
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -273,6 +273,11 @@ rh-marketplace:
         :name: API Token
         :fields:
           - :component: text-field
+            :name: authentication.authtype
+            :hideField: true
+            :initializeOnMount: true
+            :initialValue: token
+          - :component: text-field
             :name: authentication.password
             :label: API Token
             :isRequired: true
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -277,7 +277,7 @@ rh-marketplace:
             :label: API Token
             :isRequired: true
             :validate:
-              - :required:
+              - :type: required
 satellite:
   :product_name: Red Hat Satellite
   :vendor: Red Hat
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -273,7 +273,7 @@ rh-marketplace:
         :name: API Token
         :fields:
           - :component: text-field
-            :name: application.extra.api_token
+            :name: authentication.password
             :label: API Token
             :isRequired: true
             :validate:
--- a/db/seeds/source_types.yml
+++ b/db/seeds/source_types.yml
@@ -264,6 +264,20 @@ openshift:
           :is: true
   :vendor: Red Hat
   :icon_url: "/apps/frontend-assets/platform-logos/openshift-container-platform.svg"
+rh-marketplace:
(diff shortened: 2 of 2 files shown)


db/seeds/source_types.yml
```diff
@@ -264,6 +264,20 @@ openshift:
...
+rh-marketplace:
+  :product_name: Red Hat Marketplace
+  :vendor: Red Hat
+  :schema:
+    :authentication:
+      - :type: token
+        :name: API Token
+        :fields:
+          - :component: text-field
+            :name: application.extra.api_token
```
rvsia, line 276: Shouldn’t we store the token as a password (so it stays encrypted) in the authentication object? Cc @lindgrenj6

db/seeds/source_types.yml
```diff
@@ -264,6 +264,20 @@ openshift:
...
+            :label: API Token
+            :isRequired: true
+            :validate:
+              - :required:
```
rvsia, line 280: Not sure how this yaml is being translated to json, but I think it should look like this: `        - :type: required `

the result should like this in JSON format:
`validate: [{ type: “required” }]`

rvsia: If you are interested, here is a documentation for the schema we are using here: https://data-driven-forms.org/schema/introduction

db/seeds/application_types.yml
```diff
@@ -24,6 +24,14 @@
     - project_id_service_account_json
     openshift:
     - token
+"/insights/platform/rh-marketplace":
```
rvsia, line 27: I am not sure if we need any new application type.

I guess we are going to just create an source type of red hat marketplace with a token and other applications could connect to that. 🤔 

But it’s a totally new case for Sources so I am not really sure.

db/seeds/source_types.yml
rvsia: There is also a sources wizard guide: https://github.com/RedHatInsights/sources-ui/blob/master/doc/update-wizard.md

db/seeds/source_types.yml
MikelAlejoBR: You are right, yes. Thanks!

db/seeds/source_types.yml
MikelAlejoBR: Fixed. Thank you for the links!

db/seeds/application_types.yml
MikelAlejoBR: I am betting you're right on this as well. Let me ping @syncrou and @lindgrenj6 on this just to confirm it but for now I'm going to remove that part.

MikelAlejoBR:
Thank you for the guidance and help @rvsia!

db/seeds/source_types.yml
```diff
@@ -264,6 +264,20 @@ openshift:
...
   :vendor: Red Hat
   :icon_url: "/apps/frontend-assets/platform-logos/openshift-container-platform.svg"
+rh-marketplace:
+  :product_name: Red Hat Marketplace
+  :vendor: Red Hat
+  :schema:
+    :authentication:
+      - :type: token
+        :name: API Token
+        :fields:
```
rvsia, line 274: we should also add `authentication.authtype` field, check the other types

db/seeds/source_types.yml
```diff
@@ -264,6 +264,20 @@ openshift:
           :is: true
   :vendor: Red Hat
   :icon_url: "/apps/frontend-assets/platform-logos/openshift-container-platform.svg"
+rh-marketplace:
+  :product_name: Red Hat Marketplace
+  :vendor: Red Hat
```
rvsia, line 269: Currently, there are two categories in the ui: `Red Hat` and `Cloud` sources. 

Red Hat Source: `vendor == Red Hat`
Cloud Source: `vendor != Red Hat`

With introduction of Marketplace sources this is no longer possible to use - I think we could introduce a new attribute, something like `category`, to distinguish them.

cc @lindgrenj6

rvsia: (Maybe `type` would be better word.)

rvsia: So do you mean to have `Vendor` **:** `Cloud`, `Red Hat` or `Marketplace` ? 🤔 

It could work, but it is a breaking change if anyone is using the information now and we would need at least to coordinate UI/API deploy because it filters by vendor to show selection on Cloud/Red Hat section of the app.

MikelAlejoBR: Okay... so I'm unsure what to do here. Do we finally change `vendor` to `type`, or do we leave it as it is? For now it seems that the user will be able to add a single Marketplace source for the API token, and that's it...

lindgrenj6: So for now I'm thinking we should just leave it as Vendor - I know with marketplace it _technically_ isn't a Red Hat source but it can live there for now. 

We will also be needing to add certain logic for how we handle the marketplace apikey authentications (validations for only one per user, making sure that user has the right perms, etc) since they're a completely different usecase than we've had before. We can do all of that in a followup since its going to be a breaking change.

rvsia: Yeah, we can go on and merge it. It should not be a blocker, unless we are going to prod. It could get messy in the UI. 😅 

I guess we should update the disabled source types array to hide it in prod (possibly in stage too? 🤔 )

rvsia (approved):

lindgrenj6 (approved):

PR #471, opened 2022-01-04
Pull out system-specific logic from pundit checks

lindgrenj6:
Just reverting all this stuff on the API side since this is giving us grief on prod. Will implement the specific changes on the Go side that the satellite team wants.

### DIFF
--- a/app/controllers/api/v1/mixins/index_mixin.rb
+++ b/app/controllers/api/v1/mixins/index_mixin.rb
@@ -3,11 +3,9 @@ module Api
     module Mixins
       module IndexMixin
         def index
-          authorize(filtered.new)
-
           raise_unless_primary_instance_exists
           render :json => Insights::API::Common::PaginatedResponse.new(
-            :base_query => scoped(filtered.where(params_for_list.merge!(limited_access))),
+            :base_query => scoped(filtered.where(params_for_list)),
             :request    => request,
             :limit      => pagination_limit,
             :offset     => pagination_offset,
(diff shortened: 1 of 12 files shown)


syncrou:
@lindgrenj6 this is the revert of all the PSK changes correct? ie we're now going to just do "all" the PSK stuff on the GO side wih the use of the new middleware?

lindgrenj6:
Not quite - this is just pulling out all of the specific stuff related to the
`system` header authorization changes. I had went through and locked down the
system header to only POST requests as well as only viewing the things
pertaining to that certaian cert cn as much as I could - but it caused problems
for multiple clients (cost + satellite) and it was really hard to implement due
to pundit. 

PSK authorization will still be there on this side, the main change is that the
system key is still going to be treated as "auth'd". 

I'm going to go back and implement that check in Go middleware though yes - and
have a separate handler for the special satellite usecase, either that or just a
switch that changes the query a bit specifically for satellite's usecase. 

-- 
Jacob Lindgren

Red Hat
***@***.***     M: 402-841-6173

lpichler:
@lindgrenj6 there is weird CI failure but It looks there are related failures:

https://app.travis-ci.com/github/RedHatInsights/sources-api/jobs/554260559#L1616

lindgrenj6:
@lpichler Whooops - forgot to remove the `org_admin` fallback that was removed earlier. I removed that and those are gone since that is the expected behavior now.

lindgrenj6:
@lpichler can you re-review? I have it working now.

PR #479, opened 2022-01-31
Add Gemfile.lock into container

lindgrenj6:
I forgot to copy `Gemfile.lock` into the container before running `bundle install`.

`i18n` v1.9.0 also got yanked so had to run bundle update as well.


lindgrenj6:
cc @lpichler i forgot to add this - going to merge myself to fix stage.

PR #481, opened 2022-02-09
Add migration for creating rhc_connections table + join table for m2m sources relationship

lindgrenj6:


db/migrate/20220209165604_add_rhc_connections.rb
```diff
@@ -0,0 +1,25 @@
+class AddRhcConnections < ActiveRecord::Migration[5.2]
+  def up
+    create_table :rhc_connections do |t|
+      t.string :rhc_id
+      t.jsonb :extra, :default => '{}'
+      t.string :availability_status
+      t.string :availability_status_error
```
lpichler, line 7: do we need also  `last_checked_at` or `last_available_at` ?

lindgrenj6: ah yeah completely spaced those. will add them.

lpichler (approved):

PR #492, opened 2022-03-24
tmp: bypass RBAC option enabled on ClowdApp

MikelAlejoBR:
Bypass RBAC temporarily so that the performance team can set up the cluster without worrying about RBAC.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -146,6 +146,8 @@ objects:
           - name: SKIP_SEEDING
             value: ${SKIP_SEEDING}
         env:
+        - name: BYPASS_RBAC
+          value: true
         - name: APP_NAME
           value: ${APP_NAME}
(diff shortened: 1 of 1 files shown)


jhutar:
Could we maybe use:
```
        - name: BYPASS_RBAC
          value: ${BYPASS_RBAC}
```
and then later below:
```
parameters:
- description: Should we bypass RBAC option enabled? Mostly you want to keep this false.
  displayName: Bypass RBAC option enabled
  name: BYPASS_RBAC
  value: false
```
And then in app-interface we can overwrite this to `true` in our cluster?

lindgrenj6:
Yeah I agree - lets add it with a default false value in the template.

MikelAlejoBR:
Closing this PR in favor of a more meaningful one: https://github.com/RedHatInsights/sources-api/pull/493

PR #493, opened 2022-03-24
RBAC bypasseable via env var

MikelAlejoBR:
Makes it easier for other teams to enable or disable authenticating the
requests with RBAC if they deploy our application via AppInterface.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -347,7 +347,7 @@ parameters:
 - description: Skip the RBAC service entirely. All the identified requests will be assumed to be valid.
   displayName: Bypass RBAC option enabled
   name: BYPASS_RBAC
-  value: false
+  value: "false"
 - name: SECRET_KEY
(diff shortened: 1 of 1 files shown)


deploy/clowdapp.yaml
```diff
@@ -342,6 +344,10 @@ parameters:
   name: RBAC_SCHEME
   required: true
   value: http
+- description: Skip the RBAC service entirely. All the identified requests will be assumed to be valid.
+  displayName: Bypass RBAC option enabled
+  name: BYPASS_RBAC
+  value: false
```
lindgrenj6, line 350: might need to make this the string `"false"` just in case of yaml wonkiness.

MikelAlejoBR: Whoops, yeah. I saw that in the PR check but I forgot about it :hand_over_mouth:

lindgrenj6 (approved):

PR #504, opened 2022-04-26
change go password_hash migration to check for empty string (blank) passwords

lindgrenj6:
I'm already looking where password is not nil or not "" but maybe there are some in the db with empty strings.


syncrou (approved):
LGTM 👍

syncrou (approved):
LGTM 👍

